PR #2, opened 2020-03-14
Update README: Add Notes from call, streamline thinking

SomeMoosery:
## Changes:
- Added my notes from the Zoom call we had last night at 6:45. Not much to add, just a few additional notes and things that stood out to me, primarily.
- Added some additional notes and styling in order to help guide us to deciding on one critical feature of what we want our IPR to be Monday / early next week. It would be awesome if we could add multiple features by then (it seems like the symptom-tracker and a crowdsourced resource availability mechanism are both popular ideas), but I think we should focus in on at least having one core functionality done and finished before adding additional features/ideas.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,90 +1,85 @@
 # Corona-tracker
-A registry to help doctors track the symptoms of large amounts of patients real time with a dashboard.
-
+A tool to both help doctors track triage an exponentially-growing number of patients in real-time, and lead individuals to essentials they may need.
 
 ## Abstract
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -17,8 +16,7 @@ The idea is to help doctors monitor large numbers of patient's symptoms over tim
 
 We will scope it out to creating a dashboard and a survey that people update daily with information.
 
-In a few words or less, this will *"track and reduce symptoms of COVID-19."*
-
+In a few words or less, this will *"track symptoms of COVID-19 to reduce doctor overhead."*
```
SomeMoosery: I made just a bit of a change to our few-words-or-less goal, let me know if anyone has any suggestions here. I personally felt that to be true to what we're building, we ourselves won't effectively reduce any symptoms (as we aren't medical professionals and don't want to come across as experts as was stressed in that call). But, we will allow doctors to more effectively treat an ever-growing number of cases and people thinking they are showing symptoms / might be sick.

tesla809: Good points. I like the edits.

Issue #12, opened 2020-03-14
Establish connection with Blockstack


SomeMoosery:
On the frontend, we want to establish a connection with Blockstack.

This requires a little bit of discovery on how to use Blockstack and what's necessary for our app, but for now just a working POC should be good.

PR #31, opened 2020-03-15
Incorporate Blockstack Authentication

SomeMoosery:
[Ticket](https://github.com/COVID-19-electronic-health-system/Corona-tracker/issues/12)

**NOTE: PLEASE TRY THIS BEFORE MERGING. EVERYTHING WORKS FINE ON DESKTOP, BUT WHEN INSPECTING AND GOING TO THE IPHONE X PRESET, FOR EXAMPLE, YOU HAVE TO RE-SIGN IN WITH YOUR BLOCKSTACK ACCOUNT IN ORDER TO ACCESS (REQUIRES THE SECRET PHRASE AND A QUICK PASSWORD RESET (DON'T HAVE TO GO TO EMAIL TO CONFIRM). ONCE YOU'RE LOGGED IN ONCE IN MOBILE, IT WILL REMEMBER AND YOU WON'T HAVE TO RE-SIGN-IN EACH TIME. NOT A BIG DEAL FOR NOW, BUT COULD BE CUMBERSOME LATER. THIS IS A BLOCKSTACK-WIDE ISSUE... MAY BE SOMETHING TO DISCUSS LATER.**

**WHEN TESTING, LET ME KNOW IF YOU WANT THE SECRET PHRASE FOR THE BLOCKSTACK ID CORONATRACKER.ID.BLOCKSTACK I MADE**

As per the README, we decided to utilize Blockstack to create a serverless backend/database for both easy scalability (no servers to manage) and added privacy (users own their own credentials).

## Changes:
- Added the `blockstack` package. This comes with built-in functionality for the user to sign in with their Blockstack credentials (or make an account if they don't have one already). **NOTE: I've created a Blockstack account and Gmail for this project, which I can share with the group**. 
- Added `react-app-rewired`. This allows us to override the create-react-app webpack config without ejecting. This is needed to deal with CORS issues inherently caused by Blockstack (part of the Blockstack setup if using create-react-app is to either use this or to run `npm run eject`, which I'm personally not a fan of since it's irreversible). This is reflected in `config-overrides.js`. 
- Changed `App.js` to be functional. Unfortunately, unless we want to use a third-party package or some hacky workaround, Blockstack relies upon a class component for the component that implements its login.
- Added `package-lock.json` to `.gitignore`. 

### My next step will be to incorporate the UI for login, so it'll only look this ugly for a little while

### DIFF
--- a/client/config-overrides.js
+++ b/client/config-overrides.js
@@ -1,0 +1,34 @@
+module.exports = {
+    webpack: function (config, env) {
+        return config;
+    },
+    jest: function (config) {
+        return config;
+    },
+    // configFunction is the original react-scripts function that creates the
+    // Webpack Dev Server config based on the settings for proxy/allowedHost.
+    // react-scripts injects this into your function (so you can use it to
+    // create the standard config to start from), and needs to receive back a
+    // function that takes the same arguments as the original react-scripts
+    // function so that it can be used as a replacement for the original one.
+    devServer: function (configFunction) {
+        return function (proxy, allowedHost) {
+            const config = configFunction(proxy, allowedHost);
+            // Edit config here - example: set your own certificates.
+            //
+            // const fs = require('fs');
+            // config.https = {
+            //   key: fs.readFileSync(process.env.REACT_HTTPS_KEY, 'utf8'),
+            //   cert: fs.readFileSync(process.env.REACT_HTTPS_CERT, 'utf8'),
+            //   ca: fs.readFileSync(process.env.REACT_HTTPS_CA, 'utf8'),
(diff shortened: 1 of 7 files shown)


tesla809:
Question: how come App.js is changed to functional and not a class?

SomeMoosery:
So in my PR I changed `App.js` from a functional component (I know this is the more modern approach) to a class component, not class to functional. I changed this due to the nature of Blockstack's requirements to work. More details are in the details of my PR.

tesla809:
Excellent! I misread. Thanks for the clarification.

PR #34, opened 2020-03-15
Add bootstrap dependency

sedmo:


SomeMoosery:
So I've never used yarn before - we still need to import bootstrap in either `index.js` or `App.js` right? Might be worthwhile doing in this PR as well just to set a baseline

SomeMoosery:
Also be sure to rebase off master, since it looks like you have merge conflicts now. 

If you need any help with that lmk!

sedmo:
I did base off master. it seems the issue is from the one line I added in the package.json

tesla809:
We can add bootstrap dependency similar to this article. 
https://www.techiediaries.com/react-bootstrap/

So `yarn add boostrap`

Then in `src/index.js` add  `import 'bootstrap/dist/css/bootstrap.min.css';`

sedmo:
PR #35 will take place for this. Closing

PR #41, opened 2020-03-16
Theme palettehotfix

sedmo:
the color and background properties was duplicated for themePalette.css so I created seperate classes for background , slider , and text


tesla809:
I pair programmed with you on this and looks good. Merged.

PR #46, opened 2020-03-16
Navbar

dinadeljanin:


tesla809:
Looking good. However, needs to use and import the themePalete.css file. Then use the classes from it.

tesla809:
Great work @dinadeljanin !!!

dinadeljanin:
image

PR #64, opened 2020-03-17
charts components

pavel-ilin:
TempChart.js - chart component built with react-charts library
ChartJs.js - chart component built with chart.js library
I did't add this components to the App.js for this commit in order to avoid unnecessary conflicts.
Before start the server make sure that you run 'npm install' to install chart.js library.


akilhylton:
Excellent work, chart looks beautiful!!

PR #67, opened 2020-03-17
refactor ChartJs component into a functional component. Add coments t…

pavel-ilin:
1) ChartJs.js component being refactor into functional component. 
2) Add comments to ChartJs.js and ReactCharts.js

Note on the charts libraries: I used react-charts first because it was already installed, but documentation is very poor and sometimes I can't find a way to customize things there. And developers admit that this is just a beta version of their project. Instead install charts.js and build ChartJs.js component. Chart.js library allows to build more customization. I would recommend used it (or D3) instead of react-charts library.


client/src/components/App.js
```diff
@@ -59,13 +38,9 @@ export default class App extends Component {
...
-            <NavBar />
+          : <div>
             <Button onClick={this.handleSignOut}>Sign Out</Button>
-            <div>
-              <p>Welcome, { userSession.loadUserData().profile.name}</p>
-            </div>
+            <NavBar />
           </div>
         }
       </div>
```
SomeMoosery: Did you intend to edit up `App.js` here? If not, could you rebase off master if possible in order to make sure you're only touching the scope of this PR?

client/src/components/App.js
```diff
@@ -1,55 +1,34 @@
 import React, { Component } from 'react';
 import NavBar from './NavBar';
+import ChartJs from './ChartJs';
```
akilhylton: Is both ChartJs and ReactCharts files needed? If not can you remove one and keep one please.

client/src/components/App.js
pavel-ilin: I used App.js to render the charts to see how they looks like. I see that I change order of the DOM elements compared to the master.
It's been rebased.

client/src/components/App.js
pavel-ilin: I would use ChartJs because chart.js library is much customizable then react-charts.

client/src/components/App.js
SomeMoosery: Hrm, not sure the issue as I'm not super familiar with working with forked branches, but it looks like there is a lot more changed here - seems like you're still on an older version of master (if you compare current master to your master, you'll see there's no handling of user data in your `App.js` at least according to this PR). Let me know if anything needs clearing up!

SomeMoosery: Looking good now, thanks!

PR #68, opened 2020-03-17
Master translation doc from wireframes

salvolpe:
I've put in a bit of an example from the frame that I am working on in Italian. I will get more time to work on this later, but if people want to start using this format for the other languages, go right ahead!


SomeMoosery:
Great to see a translating the app! I'm going to go ahead and merge this PR so that the beginning of the template is easily-visible in the README - if you have any more changes later feel free to open up another PR

Issue #71, opened 2020-03-18
Unhandled Rejection (LoginFailedError): Failed to login: Existing user session found


BrianHHough:
<img width="1047" alt="Screen Shot 2020-03-17 at 9 20 31 PM" src="https://user-images.githubusercontent.com/56947075/76915933-54c78880-6895-11ea-9ab3-427a4af601f4.png">

Hey @SomeMoosery, i created a blockstack id, it was registered successfully, then i tried login with the login and got this error. Just wanted to flag. Everything else is looking great though!

SomeMoosery:
This is fixed

PR #72, opened 2020-03-18
HOTFIX: Fix inability to sign in

SomeMoosery:
I believe this fixes the ability where sometimes signing in would give an error. Blockstack is finicky and a lot of the documentation is a bit out of date.

@BrianHHough hopefully this should fix your [bug](https://github.com/COVID-19-electronic-health-system/Corona-tracker/issues/71)


SomeMoosery:
Merging this - if people are still having trouble signing in, please open a new issue!

PR #74, opened 2020-03-18
Update Translations.md

salvolpe:


SomeMoosery:
Looks good! Do you want me to leave this PR open so you can keep committing here if you're going to continue translating? Or would you prefer I merge and you make another PR if need be?

salvolpe:
I think that it would be easier if you could leave it open so I can continue to commit to here since we're gonna be doing work on this for a while.

PR #76, opened 2020-03-18
Setup basic Redux. Test store data avaliability in ChartJs.js component

pavel-ilin:
**Setup basic Redux:**

- Created folder "redux" in the root directory of the app
- Created rootReducer.js and add action example
- Created actions.js and add action function example
- Created store in the index.js and wrap App.js in Provider component
- Added dummy data to the store and access it through hooks in the ChartJs.js component

Now data in the redux store available in every place of the app.

**Next steps:**
- Create structure of the data and additional reducers if needed
- Fetch actual data from backend or API and update the store
- Create and dispatch actions if needed


tesla809:
Almost there! Great try!

Check the [details of the integration](https://travis-ci.org/github/COVID-19-electronic-health-system/Corona-tracker/builds/664032399?utm_source=github_status&utm_medium=notification). It has failed. 

There is an error in the code.

According to the Travis-ci, the error seems to be: 
      Error: Uncaught [Error: could not find react-redux context value; please ensure the component is wrapped in a <Provider>]

Also, would you be able to create separate /actions and /reducers folders inside of /redux? That way to new devs, their purpose might be clearer. 

Keep at it! You are doing great!

client/src/components/ChartJs.js
```diff
@@ -1,56 +1,53 @@
...
-            ticks: {
-                suggestedMax: 150,
-                suggestedMin: 0
-                    }
-                }]
-            }
-        }
-};
+// Extratring dummy data from the store 
+const dataSample = useSelector(state => state.dataSample);
```
akilhylton: Not familiar with react-redux however seem to be an error in the ChartJs component. Here is some of the error, `Error: Uncaught [Error: could not find react-redux context value; please ensure the component is wrapped in a <Provider>]`

Link to complete build error:
https://travis-ci.org/github/COVID-19-electronic-health-system/Corona-tracker/builds/664043124?utm_source=github_status&utm_medium=notification

pavel-ilin: In my understanding this error caused by lack of tests written for redux. In current test ChartJs did not wrap in Provider. It means when test tryin to render component there is no connection to redux store. And I try to use useSelector (redux hook) it fails because redux not exists in the context of this test.

Next step will be write redux test for this component.

pavel-ilin:
setupTests.js - add basic config of the enzyme testing environment
ChartJs.test.js add test with using jest and enzyme

**Next steps:**
Need to resolve: 'Failed to create chart: can't acquire context from the given item'.

PR #85, opened 2020-03-19
Initial commit for the temperature

dinadeljanin:
Barebones component for the temperature. 

Some things to note:
- Positioning will be applied once it's inside a diagnostic component, but for right now,  it sits at the top left by default.
- Font and container size does adjust for different mobile sizes.
- Left a placeholder for the temperature.
- Working on translations as well.


client/src/components/Temperature.js
```diff
@@ -0,0 +1,23 @@
...
+      <Card>
+        <Card.Body>
+            <Card.Title id="Temperature-temp">
+              98.8F
+            </Card.Title>
+            <Card.Subtitle id="Temperature-avgTempText">
+              Average Temp
+            </Card.Subtitle>
+        </Card.Body>
+      </Card>
```
SomeMoosery: Unless I'm missing something, do we necessarily need a `Card` here? I like the fact that the temperature and text "Average Temp" are tightly coupled like this, but I'm wondering if there's a way to get rid of the border? I feel that in the actual UI, that could look a little out of place.

client/src/css/NavBar.css
```diff
@@ -83,7 +83,6 @@
 @media only screen and (max-width: 414px) {
   .nav {
     height: 12vh;
-    max-width: 414px;
```
SomeMoosery: Any reason you removed this in the NavBar?

client/src/components/Temperature.js
dinadeljanin: I can do that-- I wanted to use as much react-bootstrap as i could before i made it a div, but i could switch it over to a div. 

I did notice the rounded border for the card. I wonder-- and this is going much further into the design-- perhaps even a shadow could be added to the card border itself to give it some depth.

client/src/css/NavBar.css
dinadeljanin: I remember an issue with the navbar extending far past the viewport width, and that's why that was there. I forgot to remove it in the earlier commit.

client/src/components/Temperature.js
SomeMoosery: I'll defer that question to @BrianHHough as he's been on top of the design aspect of this, but yeah I'd say either removing the border entirely or adding a showing would be the two routes here 👍

client/src/css/NavBar.css
SomeMoosery: Makes sense, Feel free to resolve this when you push your next commit removing it!

client/src/components/Temperature.js
dinadeljanin: @SomeMoosery Hey I removed the border and border radius!

SomeMoosery: Great!

PR #96, opened 2020-03-19
Fix login

friedger:
This PR
* changes the ordering of how the login state is handled (first `isUserLoggedIn` then `isLoginPending`)

### DIFF
--- a/client/src/components/App.js
+++ b/client/src/components/App.js
@@ -1,70 +1,72 @@
-import React, { Component } from 'react';
-import NavBar from './NavBar';
-import Temperature from './Temperature'
-import Login from './Login';
-import 'bootstrap/dist/css/bootstrap.min.css';
-import '../css/App.css';
-import { appConfig } from '../utils/constants'
-import { UserSession } from 'blockstack';
-import Button from 'react-bootstrap/Button';
-import { configure, User, getConfig } from 'radiks';
-import Patient from '../models/patient'
+import React, { Component } from "react";
+import NavBar from "./NavBar";
+import Temperature from "./Temperature";
+import Login from "./Login";
+import "bootstrap/dist/css/bootstrap.min.css";
+import "../css/App.css";
+import { appConfig } from "../utils/constants";
+import { UserSession } from "blockstack";
+import Button from "react-bootstrap/Button";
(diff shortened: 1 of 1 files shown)


client/src/components/App.js
```diff
@@ -1,70 +1,72 @@
...
       // Delete the entry, to keep things clean for the purpose of example
       var p;
       for (p of allPatients) {
-        p.destroy()
+        p.destroy();
       }
+    } else if (userSession.isSignInPending()) {
+      await userSession.handlePendingSignIn();
+      await User.createWithCurrentUser();
+      window.location = "/";
```
SomeMoosery: I'm just curious because maybe you've worked with Blockstack longer than I have (see: < 1 week for me) - is there any strict reasoning to put `isSignInPending()` after checking `isUserSignedIn()`? Outside of maybe slightly faster performance (which is always a plus), is it more just a code clarity thing?

client/src/components/App.js
```diff
@@ -1,70 +1,72 @@
...
-    super(props)
-    this.userSession = new UserSession({ appConfig })
+    super(props);
+    this.userSession = new UserSession({ appConfig });
   }
 
   async componentDidMount() {
+    if (!process.env.REACT_APP_QA_URL) {
+      return;
+    }
```
SomeMoosery: I wonder if, since we're just building this out for the time being to connect to the QA database just for devs, if we should throw some kind of `alert` before returning reminding devs to be sure to set their environment variables?

tesla809: Let's throw the alert() to remind devs. SInce they aren't familiar too much with codebase, it would be a helpful reminder.

client/src/components/App.js
friedger: it is about the way that `isSignInPending` is resolved: the function just looks at the url query for `authResponse`.
Therefore, it is possible the the user is signed in and sign in is pending if the url query parameter `authResponse` is not removed.

client/src/components/App.js
friedger: added

client/src/components/App.js
SomeMoosery: Ah, that makes a lot of sense - thanks!

Issue #44, opened 2020-03-16
Discuss Deployment Methodology


SomeMoosery:
So I've had a few ideas regarding this, and I'm adding this issue as a Long-Term Change for when the time comes. At a certain point, we'll want to bring this application live, and need a good way to do that. I'm a fan of AWS and have my Certified Solutions Architect certificate (although I got it a while back, and am definitely not super experience with building distributed systems from scratch myself). But, I am relatively comfy with AWS.

I've been thinking of a general architecture that we could use, which I'll attach to this issue when I draw it out. Just adding this here for a placeholder/discussion issue for the time being.

tesla809:
Amazon sounds good to me. They have great support and will probably donate credits to us. And if you are certified to do it, then let's go with that. Easier to deploy.

PR #106, opened 2020-03-20
Added information about our potential end-state architecture

SomeMoosery:
Added some architecture diagrams for what our high-level end state architecture may look like once we're ready to deploy, as well as our QA environment as well.

Updated README.

No testing required. 

Screen Shot 2020-03-19 at 10 28 45 PM

Potential end-state. Open for discussions.


SomeMoosery:
`closes #44`

friedger:
Please add
* gaia bucket of user
* app hosting server

Try to make AWS as small as possible

It could be also possible to replicate radiks and have it on different servers that we do not control.

SomeMoosery:
Thanks for the feedback @friedger - appreciate the insight, and that useful diagram! Just a couple questions before I go in to adjust. 
- What do you mean by app-hosting server? Like a more fleshed-out way of how we'd host the React app we're building? If so, I definitely agree that I should flesh that out. If that's not what you mean, I'm not sure what I'm missing - since I have the Radiks server and any backend services unaffiliated with user data (thus Radiks) accounted for here.
- When you say have Radiks on servers we don't control, do you mean host it on something [like Webtask](https://medium.com/the-lead/tutorial-running-a-blockstack-radiks-server-via-webtask-95b12e7c1a69) for example?

friedger:
@SomeMoosery 
1) I meant the hosting service where the browser is getting the code for the client code. As you mentioned it could be anywhere and the image should make it clear that  the client is not talking to the backend locally. I would draw the react app as a) a running instance on the client/browser and b) a (not-running) version on the web hosting server
2) The project could run a radiks server, but I as a user could run my own radiks server as well, indexing the public data of all users of corona tracker. The two radiks instances could collaborate for indexing, some kind of federation (not yet implemented)

friedger:
In the drawing in https://github.com/COVID-19-electronic-health-system/Corona-tracker/pull/106#issuecomment-601797896 I would put the arrow from gaia to the browser/client, not to blockstack auth.

SomeMoosery:
Closing this for now. For the time being, since we no longer plan on aggregating user data for some time, we also no longer have a need for Radiks. We've also ideated some additional microservices, such as a notifications service and an educational service, which will not be reliant on Blockstack.

I'll open a new PR with more pertinent changes when the time arises.

Issue #108, opened 2020-03-20
Add instructions for radiks server


friedger:
## Summary
The application relies on radiks server. Radiks server is using mongo db to store indexed user data.
Add information how to setup a server for development and for production

## Motivation
The repo should contain all information so that a new developer can start contributing and a new admin can start building production environment. Setup for Windows would be helpful for members of this community effort.

## Describe alternatives you've considered
There is no alternative :-)

## Additional context
https://discordapp.com/channels/687791378075484161/689541363254951951/690283491249619074

friedger:
closed by #109

friedger:
windows instructions are missing.

PR #109, opened 2020-03-20
Add minimal customized radiks server

friedger:
This PR
* adds a folder for the server part; currently, this is just an express server with radiks connecting to mongo db
* is motivated by #108 

**Test plan (required)**
1. start mongo daemon (should listen on the default port 27017)
1. checkout this branch
1. `cd server`
1. `yarn` (or `npm install`)
1. `yarn start`

Screenshot from 2020-03-20 08-49-56


SomeMoosery:
This just abstracts away the need for the user to run their own Radiks server locally, right? 

Just want to make sure I'm not missing any context

friedger:
@SomeMoosery It just prepares use for later when we want to query the radiks server for more data.
It also helps new developers to start radiks locally by just using yarn.

This is not about running a QA radiks server that we all share

SomeMoosery:
Makes sense - just wanted to be clear! Looks good otherwise, merging now 👍

PR #112, opened 2020-03-20
Issue#80

AdhamAH:
Two Things would like to note:

1- The logo and text-logo included but I think it should be in a permanent layout as it is showing everywhere
2- I don't know how will the state be defined but I tried to make it with class component and state, maybe it will be easier to modify later.

### DIFF
--- a/client/src/components/Table.js
+++ b/client/src/components/Table.js
@@ -1,61 +1,132 @@
 import React from "react";
-import'../css/Table.css'
+import "../css/Table.css";
 import "../css/themePalette.css";
 import { Table, Tabs, Tab } from "react-bootstrap";
+
+
 import { ReactComponent as Logo } from "../img/Logo_CORONATRACKER_Logo.svg";
 import { ReactComponent as TextLogo } from "../img/Logo_CORONATRACKER_Text_Logo.svg";
 
-class LogTable extends React.Component {
-    constructor() {
-        super()
-        const options = {
-            weekday: "long",
-            year: "numeric",
-            month: "long",
-            day: "numeric"
-        };
-        const today = new Date(),
-            date = today.toLocaleDateString(undefined, options)
-            
-        this.state = {
-            date: date,
-            key: ''
-        }
-    }
-    
-    render() {
-        return (
-          <div>
-            <Logo className="table-logo" />
-            <TextLogo className="table-text" />
-            <h3>
-              Hello, <b>Anthony</b>
-            </h3>
-            <h4>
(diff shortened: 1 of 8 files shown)


client/src/components/Table.js
```diff
@@ -0,0 +1,132 @@
...
+
+// Show the current date, this might should be put in App.js or as a permint layout
+const options = {
+  weekday: "long",
+  year: "numeric",
+  month: "long",
+  day: "numeric"
+};
+const today = new Date(),
+  date = today.toLocaleDateString(undefined, options);
```
SomeMoosery: So as it stands now, the Date is already creeated and displayed in `DiagnosticContainer.js` - I felt it may be better to keep in the wrapper component as we might switch views but still want to display the date, and that way we won't have to replicate getting the date across components. Let me know what you think!

client/src/components/Table.js
```diff
@@ -0,0 +1,132 @@
...
+                <Table responsive className="table">
+                  <thead className="table-head">
+                    <tr>
+                      <th>Date</th>
+                      <th>How Do I Feel</th>
+                      <th>How Are My Symptoms</th>
+                      <th>Difference From Yesterdsy</th>
+                      <th>Any Noted Symptoms</th>
+                      <th>Comments</th>
+                    </tr>
```
SomeMoosery: Is there no way to implement these tags with react-bootstrap? I don't know, I'm just curious.

tesla809: can you use destructing to pull out component 
so like 
`import React, {Component } from "react";`

Then you can extend from class like:
`class LogTable extends Component { ... }`

client/src/components/Table.js
AdhamAH: I agree that it should be in one place. I was thinking if we will make a Layout component that includes navbar, sing-out button, date components like these, things that will be showing everywhere in the app. I talked to @tesla809 and I am thinking about using material-ui . I will try something and come back to you

client/src/components/Table.js
AdhamAH: If you mean as I did with <td> tags, yeah sure it can

client/src/components/Table.js
AdhamAH: I need to do some reading about "rebase" and will do what you asked

SomeMoosery: I think using the two in parallel for the time being with an iterative shift over to Material UI is a good long-term strategy! It seems to be much more feature-rich

AdhamAH:
I made a new request

PR #114, opened 2020-03-20
Create Checkup, Evaluation Models

### LINKED ISSUE #88: Create Checkup/Survey Model
In `client/models/`, we need to create a Checkup data model for each checkup or survey associated with a CoronaTracker user.

**Acceptance Critera:**
- There is a new model in `client/models/` outlining what a checkup/survey will look like in the database

**Notes:**
- May require some discovery and finalization with regards to what information we'll be gathering. This can be incremental, though - we just need a general model for now.

SomeMoosery:
closes #88 

One of the core features of this app will be collecting and storing user data after they fill out their checkup. We need a way to model this in the database.

What I propose is creating two separate models, one that will be added to a list of daily checkups (`checkup.js`) to monitor daily health, while the other will be filled out once (or bi-weekly, or some other more extended period of time) to monitor overall clinical health and determine any preemptive risks (`evaluation.js`).

**Test plan (required)**

**TODO adding integration tests for these** - @friedger do you have any suggestions on how to go about writing integration tests with Radiks? I'm having trouble thinking it through as to query any data, a user needs to be signed in to have the data associated with a user.


ngiangre:
Question, how does this relate to the survey that's come up in different places? 

I like the idea of conforming our CDM (common data model) to both a longitudinal aspect and a initial aspect. I think the two can be integrated.

@SomeMoosery This is the CDM for clinical data https://ohdsi.github.io/TheBookOfOhdsi/CommonDataModel.html

We can develop our CDM based on their ER tables. How does that sound?

friedger:
The model looks good to me.

friedger:
LGTM

friedger:
This PR only defines the model. So far, no tests required (as we trust radiks' `Model` to be correct).

Users only need a private app key, for testing we can generate new ones, like here: https://github.com/blockstack/radiks/blob/master/test/setup.js#L93

client/src/models/observation.js
```diff
@@ -0,0 +1,55 @@
+import { Model } from 'radiks';
+
+// Observation models all observational questions we ask users
+export default class Observation extends Model {
+    static className = 'Observation';
+    static schema = {
+        coronavirus: {
```
friedger: @SomeMoosery Why did you add all these information into one model? I thought there is a daily model and a non-changing model

SomeMoosery: Modeled the data more closely to models used in the medical field after a conversation with @ngiangre - to the best of my knowledge "Observation" is a more scientifically correct umbrella which contains breakdowns (as I have here) within it - and physical versus non-physical observations is a more useful to way to break out data.

What do you think? Do you see a downside in modeling the data in this way?

friedger: Users won't update their height or weight on a daily basis, therefore, finding the latest entered height or weight requires to go through all observations (backwards until the value was found)

SomeMoosery: Good point. I can create a new PR moving age, height, and any other one-time attributes into the demographics collection, as these are really just demographics

PR #115, opened 2020-03-20
Add data for average temperature

friedger:
This PR
* adds some dummy data to the temperature component

Screenshot from 2020-03-20 20-41-27


SomeMoosery:
Great. I was actually working on a new feature and was going to add some there (even though it's not best practices in terms of scope creep) - thanks for for adding this!

PR #133, opened 2020-03-22
V2: Temperature from state

friedger:
This PR
* adds initial data as `Observations`
* adds actions to save and load observations (not used yet)
* uses state data in `Temperature` component
* loads data on first rendering


SomeMoosery:
LGTM!

PR #135, opened 2020-03-22
Updated/added .gitignore files

starving-geek:
Added .DS_Store to .gitignore files to prevent macOS generate .DS_Store files from appearing in project folders.


SomeMoosery:
Great, thanks!

PR #142, opened 2020-03-22
updates to contribution.md and logo

awu556:
Here's my edits again, but to mainly contributions, and adding the CoronaTracker logo to the top of the readme


SomeMoosery:
It looks like there's just two small merge conflicts. If you want, I can go ahead and fix them and just merge them in on my end if you're okay with that - I know it's bad practice but time is very much of the essence for the time being lol.

What do you think?

awu556:
No problem! Do what you have to do. I'll link up with the rest of the Front end team working today and see where they're at progress-wise during the 8pm check in

PR #145, opened 2020-03-22
Use reducer for observations

friedger:
This PR
* uses a reducer for observations
* adds a test for observation reducer


SomeMoosery:
Going to merge this and then create my own PR for fixing test bug - just so we can get this code merged in and master up and running.

PR #148, opened 2020-03-22
Education model

starving-geek:
I created the education model based on the education tab in the Data Model/Education spreadsheet. https://docs.google.com/spreadsheets/d/1Ji0sxIqFI49EZb7Mz4F0nkSBn1JSNMXiKjFAzJC2Jso/edit?usp=sharing


ngiangre:
the fields look good to me

starving-geek:
If there are any .DS_Store files in those files where I added/modified the .gitignore file, they will still be there. They would have to be removed. The .gitignore only prevents them from being added again.

SomeMoosery:
Ah okay, makes sense. Merging now!

PR #149, opened 2020-03-22
Check grouped buttons

liadove:
Need styling
some comments are left
image


tesla809:
quick note:
1. can we have severe be the red-like color?
2. moderate yellow
3. blue-like color

Thoughts?

ngiangre:
yes to more severe more red

liadove:
something like gradient?
image

ngiangre:
yes but diverging

client/src/components/survey-view/checkbox-button/CheckboxButton.js
```diff
@@ -0,0 +1,112 @@
+import React from 'react';
+import SingleCheckboxButton from './SingleCheckboxButton';
+
+class CheckboxButton extends React.Component {
```
friedger: Could you make it a functional component use `useState` ?

liadove: like with hooks?

liadove: I can, but is this necessary?

SomeMoosery: Using functional components is typically best practices now with React - should be used as much as possible 

https://www.codeinwp.com/blog/react-best-practices/

friedger: useCallback would be useful here.

liadove: i used useReducer, that way I don't need useCallback

SomeMoosery:
Looks good for now, if we need to make tweaks once V1 is up and running we can always come back.

Merging changes

tesla809:
Amazing work @liadove !!!

friedger:
Nice!!

PR #150, opened 2020-03-23
Material ui layout

AdhamAH:
Changed the NavBar component to material-ul 
nav

Also the DiagnosticContainer.js and made the logos always next to each other as @BrianHHough requested 

gc


client/src/components/Calendar.js
```diff
@@ -15,7 +15,7 @@ class AppCalendar extends Component {
       <div>
         <Calendar
           onChange={this.onChange}
-          value={this.state.date}
+        value={this.state.date}
```
friedger: unrelated change here, please revert.

client/src/components/DiagnosticContainer.js
```diff
@@ -6,51 +6,61 @@ import { ReactComponent as Logo } from '../img/Logo_CORONATRACKER_Logo.svg';
 import { ReactComponent as TextLogo } from '../img/Logo_CORONATRACKER_Text_Logo.svg';
 import NavBar from './NavBar';
 import Button from 'react-bootstrap/Button';
-import { Container, Row, Col } from 'react-bootstrap';
-import Temperature from './Temperature';
+import { makeStyles } from "@material-ui/core/styles";
+import { Container } from "@material-ui/core";
 
+
+const useStyles = makeStyles(theme => ({
```
friedger: Was there a decision to use withStyles instead of css files?

client/src/components/NavBar.js
```diff
@@ -1,49 +1,105 @@
...
+              classes={{
+                root: classes.actionItem,
+                selected: classes.selected,
+                indicator: classes.indicator
+              }}
+              icon={<FeaturedPlayListOutlinedIcon className={classes.buttons}/>}
+            />
+            <BottomNavigationAction
+              label="Education"
+              value="Education"
```
friedger: should be probably lower case

client/src/components/Table.js
```diff
@@ -0,0 +1,108 @@
...
+import { Table } from "react-bootstrap";
+
+
+
+// The Table
+class LogTable extends Component {
+  constructor() {
+    super();
+    // Made dummy data as state becuase it might be easier to change later
+    this.state = {
```
friedger: Component state should only be used for data that is changed by the user, just use a constant here.

client/src/components/Table.js
```diff
@@ -0,0 +1,108 @@
...
+      questionnaire: [
+        "Date",
+        "How Do you Feel?",
+        "How Are My Symptoms?",
+        "Difference From Yesterdsy",
+        "Any Noted Symptoms",
+        "Comments"
+      ],
+      log: [
```
friedger: We have already the `Observation` model. You should derive the display model from these.

client/src/components/Table.js
```diff
@@ -0,0 +1,108 @@
...
+            {this.state.log.map(
+              ({
+                date,
+                feel,
+                symptoms,
+                yesterdsy,
+                notedSymptoms,
+                comments
+              }) => (
+                  <tbody className="table-body">
```
friedger: `tbody` should be outside of `map`

client/src/components/Table.js
```diff
@@ -0,0 +1,108 @@
...
+                    <tr>
```
friedger: list of items needs a `key`, something like `<tr key={date}>`

client/src/components/Temperature.js
```diff
@@ -11,8 +11,8 @@ export default ({ allRecords }) => {
 
   return (
     <Card>
-      <Card.Body>
-        <Card.Title id="Temperature-temp">{averageTemperature}F</Card.Title>
+      <Card.Body style={{width: '130px'}}>
```
friedger: Please move to a css file.

client/package.json
```diff
@@ -7,18 +7,20 @@
...
     "enzyme-adapter-react-16": "1.14.0",
+    "i": "^0.3.6",
     "jest": "24.9.0",
     "jest-enzyme": "7.1.0",
-    "react-test-renderer": "16.9.0",
-    "redux-mock-store": "1.5.3",
-    "axios": "^0.19.2",
-    "blockstack": "^19.3.0",
-    "bootstrap": "^4.4.1",
+    "npm": "^6.14.3",
```
friedger: What is `npm` for?

AdhamAH:
too many conflicting files and much change

PR #152, opened 2020-03-23
feature: added navigation with react-router and removed icon from navbar

Ahwan0717:
Navigation works. I added react-router and all of the routes without their components since they are unfinished. I added all of the path names that I thought we would need based on the final wireframes. Anyone is more than welcome to suggest more/add to it. Also removed the chat icon. 

Screen Shot 2020-03-22 at 10 53 39 PM
Screen Shot 2020-03-22 at 10 53 27 PM
Screen Shot 2020-03-22 at 10 53 06 PM


tesla809:
sweet! looks great @Ahwan0717 !

Issue #137, opened 2020-03-22
[Education]: Fact / Quiz Interactive Mode 3.22.20


awu556:
See wireframe ![Fact / Quiz Screen](https://github.com/COVID-19-electronic-health-system/Corona-tracker/blob/master/design/wireframes/v1-final-MVP/CORONATRACKER%20UI%20-%20Education%20(Fact%20Mode).png)

- [x] Component 1: Upfront logo + text-based logo

- [x] Component 2: Two-button
Left: “Fact Mode” -- make red
Right: “Quiz Mode” -- make blank

- [x] Component 3: Instructions
Left: “Fact Mode” -- make red
Right: “Quiz Mode” -- make blank

- [x] Component 4: Flash Cards
“Flash card” like visualization with cards “stacked on top of one another (maybe make it look like there’s 3-4 with the top one being locked in place once you swipe the top one away)
The card has a few elements
https://github.com/oliviertassinari/react-swipeable-views

Main goal to gameify education

johnemurphy1:
Adding these rough drafts here for the fact/quiz format button
https://github.com/johnemurphy1/Corona-tracker/blob/master/EducationTwoButton.css
https://github.com/johnemurphy1/Corona-tracker/blob/master/EducationTwoButton.js

johnemurphy1:
ok, i'm not going to do a PR on the fact/quiz button. latest version that @Tanvez posted has it working...

awu556:
Excellent work the both of you! We couldn't have reached V1 without your help.

PR #162, opened 2020-03-23
Added Flash Card and Card

Tanvez:
Need some help to do some styling to get it look like the mockups. This is for the Educational section.
ezgif-3-3afd7e14e490
ezgif-3-5236d8948ec6

Issue #137


client/src/components/Card.js
```diff
@@ -0,0 +1,40 @@
+import React from "react";
+import { animated, interpolate } from "react-spring";
+import Button from '@material-ui/core/Button';
+
+class Card extends React.Component {
```
SomeMoosery: Is there any reason this is a class component, and not functional? If not, let's try to keep things consistent as functional components as you've done with flashcards

Tanvez: Will do!

SomeMoosery: Looks good!

SomeMoosery:
Looking good so far @Tanvez - just seems like there's still pesky merge conflicts that need rebasing. 

[This is a good place to start, let me know if you need any help!](https://stackoverflow.com/questions/15602037/git-rebase-upstream-master-vs-git-pull-rebase-upstream-master)

PR #167, opened 2020-03-23
fix: cleaning up artifacts from create react app

crispinamuriel:
added favicon, changed up create-react-app <title>


SomeMoosery:
Looks good! Merging

PR #178, opened 2020-03-23
Added conditional to DiagnosticContainer to show Question 

AdhamAH:
Screenshot 2020-03-24 at 00 42 02
depending on the state
Screenshot 2020-03-24 at 00 42 26
Screenshot 2020-03-24 at 00 42 51

### DIFF
--- a/client/src/components/DiagnosticContainer.js
+++ b/client/src/components/DiagnosticContainer.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import '../css/App.css';
 import '../css/DiagnosticContainer.css';
 import '../css/themePalette.css';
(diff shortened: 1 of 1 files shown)


liadove:
Good job!

AdhamAH:
> Good job!

like the app lol

client/src/components/DiagnosticContainer.js
```diff
@@ -39,6 +43,7 @@ const dateOptions = {
 };
 
 function DiagnosticContainer(props) {
+  const [healthLogged] = useState(false);
```
friedger: This should come from redux, shouldn't it?

liadove: Yes, it will once we know what redux would look like. For now, it's just for condition underneath

SomeMoosery: We should still get state from Redux whenever possible as it should be the single source of truth for application state, even if it's just placeholder state for now

AdhamAH:
As seen from the comments, no need for it.

PR #185, opened 2020-03-24
Disclaimer popup styling

Afrogrrrl:
So far there's only logic for someone clicking agree (a bool switch on state) and nothing happens if you click disagree. Logic for agree/disagree should probably be made a separate issue

### DIFF
--- a/client/src/components/Disclaimer.js
+++ b/client/src/components/Disclaimer.js
@@ -3,43 +3,28 @@ import { ReactComponent as Logo } from '../img/Logo_CORONATRACKER_Logo.svg';
 import { ReactComponent as TextLogo } from '../img/Logo_CORONATRACKER_Text_Logo.svg';
 import Button from 'react-bootstrap/Button';
 
-class Disclaimer extends Component {
-  constructor() {
-    super();
-    this.state = {
-      showPopUp: true,
-    };
-  }
-
-  render() {
-    return (
-      <div>
-        {this.state.showPopUp ? (
-          <div className="DisclaimerPopUp">
-            <div className="DisclaimerPopUpInner">
-              <Logo className="LoginLogo" className="PopUpLogo" />
-              <TextLogo className="TextLogo" className="PopUpLogo" />
-              <p>
-                The CoronaTracker is designed to help you navigate through the pandemic with accurate information,
-                charting your wellbeing, and tracking your health. The CoronaTracker is not intended to be used or
-                viewed as diagnosis or treatment of disease or other symptoms, including but not limited to COVID-19.
-              </p>
-              <p>
-                This application was made by the community for the community. This application is made for you! You are
-                in control of how you would like to contribute your data for public health and research.
-              </p>
-              <div>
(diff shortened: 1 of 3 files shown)


client/src/components/Disclaimer.js
```diff
@@ -1,29 +1,45 @@
-import React, {Component} from 'react';
+import React, { Component } from 'react';
 import { ReactComponent as Logo } from '../img/Logo_CORONATRACKER_Logo.svg';
 import { ReactComponent as TextLogo } from '../img/Logo_CORONATRACKER_Text_Logo.svg';
 import Button from 'react-bootstrap/Button';
 
 class Disclaimer extends Component {
```
SomeMoosery: I know we already have Disclaimer created, but could you rewrite this into a functional component in this PR for consistency?

client/src/components/Disclaimer.js
```diff
@@ -1,29 +1,45 @@
...
+  constructor() {
+    super();
+    this.state = {
+      showPopUp: true,
+    };
```
SomeMoosery: I feel like a better place to handle this logic would be in `DiagnosticContainer` using Redux.

We have an initial state of `true` for `showPopUp` in Redux. Then, when rendering `DiagnosticContainer`, we determine if we need to render `Disclaimer` at all.

client/src/components/Disclaimer.js
```diff
@@ -1,29 +1,45 @@
...
+                viewed as diagnosis or treatment of disease or other symptoms, including but not limited to COVID-19.
+              </p>
+              <p>
+                This application was made by the community for the community. This application is made for you! You are
+                in control of how you would like to contribute your data for public health and research.
+              </p>
+              <div>
+                <Button className="theme-Palette-white-text" onClick={() => this.setState({ showPopUp: false })}>
+                  I agree
+                </Button>
```
SomeMoosery: Here, instead of setting the state within this component, we would set the state with Redux, and `DiagnosticContainer` would re-render and not even load this at all.

client/src/components/Disclaimer.js
```diff
@@ -1,29 +1,45 @@
...
+                <Button className="theme-Palette-white-text theme-Palette-red-bg">I don't agree</Button>
```
SomeMoosery: When we do implement "I don't agree" functionality, this will also just set the state with Redux to false, and then we just redirect to login.

client/src/components/Disclaimer.js
Afrogrrrl: sure thing!

client/src/components/Disclaimer.js
Afrogrrrl: gotcha. Should I do that for this PR or make it a separate thing?

Afrogrrrl:
made into a functional component, and I commented out from the Diagnostics Container component so it doesn't just show up without being able to make it go away.

SomeMoosery:
Awesome, great changes!

PR #186, opened 2020-03-24
Styling change cards & NavBar

### LINKED ISSUE #139: [Homepage Dashboard]: Navbar, Layout Formatting for Diagnostic Container 3.22.20
Component 1: Upfront welcome

- [x] Logo to the left + text-logo to the right all on one 
- [x] Good morning, { firstName }
- [x] Today is, { dayOfWeek, month #, year 
- [x] Small black horizontal bar
- [x] Prompt that reads, either: 
         - “Let’s log your health for today” (if empty)
         - “Great job logging  your health today (if fulfilled)
         - Only shows if you have yet to log it, maybe say something like “Great job logging your health      for today!”

AdhamAH:
Changed the style of the cards 

fixes #139

Screenshot 2020-03-24 at 19 06 54

Added. some style to NavBar to be more interactive 

Screenshot 2020-03-24 at 19 07 04

### DIFF
--- a/client/src/components/NavBar.js
+++ b/client/src/components/NavBar.js
@@ -53,6 +53,8 @@ class NavBar extends React.Component {
     return (
       <BottomNavigation className={classes.root} showLabels>
         <BottomNavigationAction
+          component={Link}
+          to='/'
           label="Log"
           value="log"
           classes={{
--- a/client/src/components/NavBar.js
+++ b/client/src/components/NavBar.js
@@ -1,17 +1,14 @@
 import React from 'react';
-import { connect } from 'react-redux';
 import '../css/themePalette.css';
 import '../css/NavBar.css';
 import { withStyles } from '@material-ui/core/styles';
-
+import { Link } from 'react-router-dom';
 import { BottomNavigation, BottomNavigationAction } from '@material-ui/core/';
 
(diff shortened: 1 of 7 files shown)


client/src/components/NavBar.js
```diff
@@ -24,32 +21,42 @@ const styles = {
...
-  // };
+  state = {
+    value: 'log',
+  };
 
-  // handleChange = (event, value) => {
-  //   console.log(value);
-  //   this.setState({ value });
-  // };
+  handleChange = (event, value) => {
```
friedger: What is the motivation to take the navigation out of redux again?

friedger: If by accident, this should be reverted. If on purpose this should go into a new PR

SomeMoosery: Was this resolved? The Files Changed I'm seeing still has this PR removing redux

SomeMoosery:
For the sake of time and getting things going before an alpha release for feedback Friday, we've decided to move forward without Redux in this component. [Going off the Redux docs](https://redux.js.org/advanced/usage-with-react-router), it seems like while Redux is the SSOT for data, Router is the SSOT for, well, routing - so it may not make sense to intermingle the two. 

Open to ideas and suggestions of course, happy to open a new issue if the need arises.

Issue #164, opened 2020-03-23
Extract private route component


friedger:
## Summary
* Move redirect to Login into a `PrivateRoute` functional component
* Replace Route with PrivateRoute for `log`, `symptomsurvey`, `healthlog`, `settings`

## Motivation
This refactoring make the routing more readable.

fallon7284:
I'm working on this.

fallon7284:
I'm curious to know whether userSession will ever be stored in the Redux store.  If it was in the store I could just connect the PrivateRoute component so we don't have to pass the userSession to each PrivateRoute.

AdhamAH:
Thanks for working on it. @pavel-ilin you worked in Redux the most if I am not wrong. Can you maybe answer Brendan

fallon7284:
I'm just going to pass it as props for now to keep the PR narrow and it will be easy enough to change later.

SomeMoosery:
Which parts of `userSession` do you need to handle? I may not have the full scope of this issue, but would it be possible to just store the parts of `userSession` you need in Redux store? Or utilize the Radiks API to get `userSession` data wherever needed?

liadove:
@fallon7284 did you solve the issue with radiks?

fallon7284:
No I still have an issue with Radiks, but it's definitely a separate issue and I'm going to PR this, just hoping someone can point me in the direction of the PR template.  The link in the readme is dead.

SomeMoosery:
Sorry about that @fallon7284 I'll get that fixed tonight. [Here's the link](https://github.com/COVID-19-electronic-health-system/Corona-tracker/blob/master/.github/ISSUE_TEMPLATE/pull_request_template.md) to the PR template!

PR #191, opened 2020-03-25
refactor: routes

liadove:
I refactored the routes part. So it will redirect you to /log or /healthlog depended on if you are logged in. And also it won't allow you to go on some routes if you don't have access to it. Probably we'll need to change callback for blackstack also since I thought it would make sense if login part be on route /log.

### DIFF
--- a/client/src/components/App.js
+++ b/client/src/components/App.js
@@ -7,7 +7,7 @@ import { UserSession } from 'blockstack';
 import { configure, User, getConfig } from 'radiks';
 import { Connect } from '@blockstack/connect';
 import DiagnosticContainer from './DiagnosticContainer';
-import { BrowserRouter, Route, Switch } from 'react-router-dom';
+import { BrowserRouter, Route, Switch, Redirect } from 'react-router-dom';
 import { connect } from 'react-redux';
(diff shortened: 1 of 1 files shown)


SomeMoosery:
This is definitely a good idea and good contribution - it's important we only show relevant data if the user is logged in - but before reviewing / moving forward this seems related to #164 - maybe link up with @fallon7284 who is actively working on that ticket to see if you can share ideas / combine this into one PR? 

Going forward I think it's best to only open PRs related to issues, and if there are no issues relating to the PR open an issue first - just to keep things organized 🚀

liadove:
Thanks for feedback! I will definitely talk to @fallon7284 and next time will open an issue or find one.

fallon7284:
Yeah so per issue #164 I made a PrivateRoute component that handles that conditional logic internal: you pass it a component as a prop and if the user is logged in it renders that component, otherwise it renders the login.  I'm just trying to work out some issues with my radiks server before I PR it because I want to make sure that is the source of an issue I'm having, rather than my code.

liadove:
Hi @fallon7284, sounds good!
So then I just close this PR.

PR #193, opened 2020-03-25
#164 PrivateRoute component

fallon7284:
**Description of change** => replaces protected routes with PrivateRoute component, which receives as props the authed state, the path, and the component to render, and conditionally renders either the component received as a prop or the Login component.

**Alternate designs** => either keep authed state in redux so that we can connect the PrivateRoute component to the store and not need to pass authed on every route, or... I've been given the impression that there is a way that we can access the userSession data from Radiks in any component, so that could provide a similar solution. I have no experience in Radiks and haven't been able to figure out how to do that.

**Possible drawbacks** => need to pass authed boolean to every route, may eventually affect readability.

### DIFF
--- a/client/src/components/App.js
+++ b/client/src/components/App.js
@@ -75,12 +75,16 @@ class App extends Component {
               />
 
               {/* ADD/EDIT ROUTES WITH THEIR COMPONENTS HERE: */}
-              <Route path="/signup" />
+              <PrivateRoute path="/signup" authed={authed} />
               <PrivateRoute path="/symptomsurvey" authed={authed} />
               <PrivateRoute path="/log" authed={authed} />
               <PrivateRoute path="/healthlog" authed={authed} />
-              <Route path="/education" render={() => <FactQuizContainer handleSignOut={this.handleSignOut} />} />
-              <Route path="/map" />
+              <PrivateRoute
+                path="/education"
+                authed={authed}
+                render={() => <FactQuizContainer handleSignOut={this.handleSignOut} />}
+              />
+              <PrivateRoute path="/map" authed={authed} />
               <PrivateRoute path="/settings" authed={authed} />
             </Switch>
           </div>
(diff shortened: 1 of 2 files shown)


client/src/components/App.js
```diff
@@ -61,29 +62,26 @@ class App extends Component {
...
               {/* ADD/EDIT ROUTES WITH THEIR COMPONENTS HERE: */}
               <Route path="/signup" />
-              <Route path="/symptomsurvey" />
-              <Route path="/log" />
-              <Route path="/healthlog" />
+              <PrivateRoute path="/symptomsurvey" authed={authed} />
+              <PrivateRoute path="/log" authed={authed} />
+              <PrivateRoute path="/healthlog" authed={authed} />
               <Route path="/education" render={() => <FactQuizContainer handleSignOut={this.handleSignOut} />} />
               <Route path="/map" />
```
liadove: should /education and /map paths be only for logged in users?

SomeMoosery: ^ Up for discussion, but I'd say yes, at least for the time being. Otherwise, we'd have to add a way to reach eduction or map before logging in. Which, now that I'm thinking about it, could be a useful feature down the line, but I don't think we should concern ourselves with it now.

fallon7284:
> does it redirect you to sign up or log tab when you type a path that doesn't exist or you don't have access to?

It does not, I tried to stick to the scope of issue #164 which was:

- Move redirect to Login into a PrivateRoute functional component
- Replace Route with PrivateRoute for log, symptomsurvey, healthlog, settings

client/src/components/App.js
fallon7284: *updated, /education and /map paths are also PrivateRoutes.

AdhamAH:
This will make it easy for an idea I have in mind to redirect if the user disagrees to other than outside the app or to the login page.

Issue #188, opened 2020-03-24
Redirect to CDC  on disagree


AdhamAH:
**This was not meant to be a question to chose between CDC and WHO. This can be done later depending on the locality of the user. Which I think is @SomeMoosery expertise from server-side. If not possible then another dialogue to choose where. Now the main initial target is the US, later we may add also RIMV (for the Dutch 😂)**
## Idea 

When a user disagrees to the Disclaimer [issue#140](https://github.com/COVID-19-electronic-health-system/Corona-tracker/issues/140), the app will redirect to CDC or WHO page 

## Motivation 
Rather than just log the user out we redirect him to another source of info. In the end, we need users to know more about the virus and the user might find it a good thing and come back to use it 

## How to do it
There are many ways but as we are using react-router, we can use Redirect in a simple conditional statement

friedger:
As we aim for global reach, redirecting to WHO would be most suitable.

Afrogrrrl:
^ seconded on redirecting to WHO. I actually think it should log someone out and redirect them to WHO

AdhamAH:
> As we aim for global reach, redirecting to WHO would be most suitable.

My bad, I didn't ask the right question
  

> ^ seconded on redirecting to WHO. I actually think it should log someone out and redirect them to WHO
thanks for pointing out the logout part.

AdhamAH:
I was waiting for the changes to be made to the disclaimer. Now it looks like it is not needed anymore

SomeMoosery:
Sounds good. We can always open another ticket later if we decide to do something differently :)

PR #197, opened 2020-03-26
Fixed routing to Education 

### LINKED ISSUE #196: Education tab not rendering component
* [ X ] Put an X between the brackets on this line if you have done all of the following:

### Description

When running the app in QA, navigating to the Education tab does not render anything. This looks to be a problem in the render parameter in the associated `PrivateRoute`.

### Steps to Reproduce

1. Run the application in QA
2. Navigate to education tab

**Expected behavior:**

Education tab loads flashcards as it was before

**Actual behavior:**

Nothing renders

**Reproduces how often:**

100%

awu556:
Changed attribute in PrivateRoute (render => component). You can now navigate to the education page. Also added state to the Disclaimer.js so when uncommented, the disclaimer will disappear after agreeing. 

fixes #196

### DIFF
--- a/client/src/components/Disclaimer.js
+++ b/client/src/components/Disclaimer.js
@@ -4,25 +4,32 @@ import { ReactComponent as TextLogo } from '../img/Logo_CORONATRACKER_Text_Logo.
 import Button from 'react-bootstrap/Button';
 
 function Disclaimer(props) {
+  const [modalHidden, setModalHidden] = React.useState(true);
   return (
-    <div className="DisclaimerPopUp">
-      <div className="DisclaimerPopUpInner">
-        <Logo className="LoginLogo" className="PopUpLogo" />
-        <TextLogo className="TextLogo" className="PopUpLogo" />
-        <p>
-          The CoronaTracker is designed to help you navigate through the pandemic with accurate information, charting
-          your wellbeing, and tracking your health. The CoronaTracker is not intended to be used or viewed as diagnosis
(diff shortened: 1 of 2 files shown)


client/src/components/Disclaimer.js
```diff
@@ -4,25 +4,32 @@ import { ReactComponent as TextLogo } from '../img/Logo_CORONATRACKER_Text_Logo.
...
+            <div>
+              <Button className="theme-Palette-white-text" onClick={() => setModalHidden(false)}>
+                I agree
+              </Button>
+              <Button className="theme-Palette-white-text theme-Palette-red-bg">I don't agree</Button>
+            </div>
+          </div>
         </div>
-      </div>
+      ) : null}
```
SomeMoosery: While this will be needed in the future, we should try to keep PRs scoped to tackling one issue at a time.

We should remove this but then should definitely open up another issue addressing making `Disclimaer.js` stateful, and when we do, we'll want to use Redux as we should try and keep it as much of a SSOT as possible!

awu556: Good point, I'll remove it for now and just add the fix for the route!

AdhamAH: I was typing haha! but @SomeMoosery. Also I wanted to point out that there is an issue [here](https://github.com/COVID-19-electronic-health-system/Corona-tracker/issues/188) and soon will be changed to material-ui

PR #206, opened 2020-03-26
HOTFIX: bug in disclaimer.js not defined setModalHidden line 23

liadove:


client/src/components/HealthLogToggle.js
```diff
@@ -16,7 +16,7 @@ export default function HealthLogToggle() {
         <Button onClick={setShowMeMore} variant={toggleValue === 'showMeMore' && "contained"}>Show me more</Button>
       </ButtonGroup>
       </Container>
-      {toggleValue === 'myHealthLog' && <SymptomsTracker />}
+      {/* {toggleValue === 'myHealthLog' && < />} */}
```
SomeMoosery: Why are we changing this/commenting it out?

client/src/components/Disclaimer.js
```diff
@@ -19,7 +19,9 @@ function Disclaimer(props) {
           control of how you would like to contribute your data for public health and research.
         </p>
         <div>
-          <Button className="theme-Palette-white-text" onClick={() => setModalHidden(false)}>
+          <Button className="theme-Palette-white-text"
+          //  onClick={() => setModalHidden(false)}
+           >
```
SomeMoosery: I think it would be better to just remove this `onClick` entirely for the time being, leaving the formatting the same on one line - that way it'll just reduce the chances of a merge conflict when we implement `setModalHidden` later tonight - what do you think?

client/src/components/HealthLogToggle.js
liadove: Ohh I deleted it sorry! We don't need it there instead we need to put myHealthLog component with the calendar

client/src/components/Disclaimer.js
liadove: Yes, sure! will update now

client/src/components/HealthLogToggle.js
SomeMoosery: Ah okay, that makes sense!

client/src/components/Disclaimer.js
SomeMoosery: Thanks!

PR #207, opened 2020-03-26
Calendar.js component build with fullcalendar library

### LINKED ISSUE #14: Create Calendar component as per mock up in design directory
**Task**
Use the mockup in the design directory to create a separate component for the calendar.

**What done looks like**
We have a Calendar.js component in /components.
We have a .css file to style the Calendar.js component in /css.

**Design considerations**
We have decided to use regular CSS files to keep things simple for now. This allows for more contributors and less time to upskill to contribute. 

Once MVP is done the team can discuss if we want to transition it to styled-components or something like CSS modules. 

**Folder paths**
CSS is located in /client/src/css
Components are located in /client/src/components

Style guide for CSS files:

**Naming convention:**
Component -> Calendar.js
CSS file -> Calendar.css

**To avoid naming collisions:**
prefix class names with the name of component and link with kebab case.
Class name -> .Calendar-header { ... }

**How to handle state:**
If you can use redux, use it. 
Create action and reducer with an appropriate name.

**testing**
No need to worry about testing for now. 
If you choose to add tests, please place in client/src/tests

pavel-ilin:
fixes #14 

Thank you @BrianHHough for checking out fullcalendar and provide access to their access without copyright restriction.

**Done**
Calendar.js component render a calendar
Calendar.css - calendar container style
package.json - add fullcalendar libraries
 
**User** Stories
- User can slide to different dates on the calendar (different years, months and days)
- User can create events on the calendar
- User can switch between month/week/day views

**Nets steps**
- Decide what calendar should do, what data should be render on it an what user stories we want to develop with calendar.


client/src/components/Calendar.js
```diff
@@ -1,25 +1,49 @@
...
 
+  // create new event function
+  const handleCreateEventClick = (arg) => {
+    let arrayObject = {
+      title: 'new event',
+      start: arg.date,
+      allDay: arg.allDay
+    }
+    setCalendarEvents([...calendarEvents, arrayObject])
+  }
```
SomeMoosery: I think using `useState` is fine for now as this Calendar is more for aesthetic purposes for now, but in the future I'm just noting that we may refactor this to implement Redux, as we might be storing some information here that might be passed through multiple components

pavel-ilin: I'm more then happy to refactor this component with useSelect when we will have necessary data points in the redux store and useDispatch action to update the store.

PR #217, opened 2020-03-26
make layout

### LINKED ISSUE #189: Add constant components in top-level component
## Idea:
To make a united layout throughout the app.

1. Include top-level components like navbar, constant images like logos in a top-level component such as `Layout.js`. Remove them from DiagnosticContainer component
2. Use Material-UI theming
3. As per @friedger and @AdhamAH suggestions, keep constant components in Layout.js’ and React providers in ‘App.js’ 
 
## Motivation:

1.  Keep consistency in the styling of the app
2. Make it easier for anyone new coming on board to follow the same flow
3. Start implementing Material-UI

AdhamAH:
fixes #189

### DIFF
--- a/client/src/components/Layout.js
+++ b/client/src/components/Layout.js
@@ -1,0 +1,93 @@
+import React from 'react'
+import { MuiThemeProvider, createMuiTheme, CssBaseline } from '@material-ui/core';
+import { makeStyles } from '@material-ui/core/styles';
+import Button from '@material-ui/core/Button';
+import { ReactComponent as Logo } from '../img/Logo_CORONATRACKER_Logo.svg';
+import { ReactComponent as TextLogo } from '../img/Logo_CORONATRACKER_Text_Logo.svg';
+import NavBar from './NavBar';
+
+import Login from './Login';
+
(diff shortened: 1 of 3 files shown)


client/src/components/Layout.js
```diff
@@ -0,0 +1,93 @@
...
+            <Button size="medium" color="secondary" variant="contained" onClick={handleSignOut}>
+              Sign Out
+            </Button>
+            <h4>
+              Hello, <b>{userSession.loadUserData().profile.name} </b>
+            </h4>
+            <h5>
+              Today is <b>{today.toLocaleDateString(undefined, dateOptions)}</b>{' '}
+            </h5>
+            <hr className="hr" />
```
liadove: Just a thought: not everywhere we need to have Hello and date part. I think it's only for tab1? Would it be reusable?

liadove: Ohh I didn't see it's merged.. I can fix it on my end if you're okay with that?

AdhamAH: Yes please, but I think you need to make an issue and then when you do the PR add the fix

PR #220, opened 2020-03-26
changed FactQuiz style to go with layout

AdhamAH:
Deleted some undeeded elements from FactQuiz as they are in the layout #217


liadove:
Awesome! Good job @AdhamAH !

PR #228, opened 2020-03-27
closes issue #205 Disclaimer saving state

### LINKED ISSUE #205: Disclaimer saving state
## Summary

Have redux variable for the user answer for disclaimer, which will store values: 'agree', 'not agree' (true, false?). In the future, we'll store it in the database inside the user table. So that way we'll know if the user agreed to the disclaimer.

## Motivation

Why are we doing this? 
We don't want the disclaimer to pop up every time and it won't go away because buttons currently don't work.

What is the expected outcome? 
Disclaimer state will save the user answer and store it redux. And depends on the answer it will send you back or send it to the main tab.

Have maybe a user profile redux state, that will have a property with a disclaimer answer (initially could be null). Then send it to the disclaimer component and component will change the answer and send to redux, and then to the backend where it will be stored, so the next we won't ask a user to agree or disagree.

Afrogrrrl:
fixes #205 

Writes 'agree' or 'disagree' to redux store on click of respective buttons on popup. Current logic has it so that the popup will only go away upon clicking agree. if user clicks disagree and goes to another page then comes back, popup still appears


client/src/redux/reducers/disclaimer.js
```diff
@@ -1,18 +1,19 @@
-import { SET_DISCLAIMER_MODAL } from '../actions/actions';
+import { DISCLAIMER_ANSWER } from '../actions/actions';
 
-const INITIAL_STATE = {
-    agreeToDisclaimer: false
-}
+const initialState = {
+    answer: null,
+};
```
SomeMoosery: Was there any benefit of moving away from a boolean agree/disagree variable? Why not just have the initial state be false, and then only change to true when the user consents?

client/src/components/Disclaimer.js
```diff
@@ -3,46 +3,59 @@ import { ReactComponent as Logo } from '../img/Logo_CORONATRACKER_Logo.svg';
 import { ReactComponent as TextLogo } from '../img/Logo_CORONATRACKER_Text_Logo.svg';
 import Button from 'react-bootstrap/Button';
 import { connect } from 'react-redux';
-import actions from '../redux/actions/actions'
-import '../css/Disclaimer.css'
+import actions from '../redux/actions/actions';
+import { getConfig } from 'radiks';
+
+export const Disclaimer = props => {
+  console.log(props);
```
liadove: Delete console.log

client/src/redux/reducers/disclaimer.js
liadove: Do we need to know if a user didn't click anything yet? What should happen when he disagrees?

Afrogrrrl: Good question. My thought process was basically that theoretically if a user logs in and clicks disagree, that'll do it's thing (decision pending on what exactly I guess, but say it includes logout) and if they log back in and the app remembers that user specifically clicked disagree, I think we would still want the popup to come up again. If it's just set to false I feel like it doesn't really differentiate from that case vs it's false because user has yet to click either option.

Afrogrrrl: So right now the logic is set like if the value is 'null' OR 'disagree', then popup again. I suppose that a bool would do just fine here, I guess I was thinking of a future case where we'd specifically want to know that the user clicked disagree.

liadove: So I am a little confused, but I understand it like if you click disagree you would automatically logout? Then if you attempt to log in again you still need to be asked to agree, right? Is there a difference between when you log in for the first time and when you try to log in after disagreeing?

Afrogrrrl: As of now there isn't a difference. You're both right, I'll just change it to a bool and if we do come across a case where we need to know that user specifically clicked disagree it can be changed then.

SomeMoosery: I don't think the click would determine when to redirect though.

The way I'm seeing it is you could keep it as a boolean value initialized to false. The user click disagree and an onclick (for example) callback redirects them to the WHO/CDC/Login/etc. This wouldn't necessarily update the state. It could still be false and the next time they open the app they'd still face this option.

I'm just not seeing a big differentiation from when they initially log into the app for the first time, or when they log back in after disagreeing once before. What do you think? Am I missing any context?

SomeMoosery: Ah, just saw your message. Sounds good!

Afrogrrrl: updated it!

SomeMoosery:
Nice job!

PR #231, opened 2020-03-27
feat: added Subscribe component Issue #224

### LINKED ISSUE #224: Create "Subscribe Field/Button for Alpha Testing"
## Summary

For the alpha test releasing Friday, one of the milestones is a basic notification service. While we don't necessarily need to decide on the final location of this, we need to present the user with the ability to subscribe to our notifications service.

For the time being, this can be a simple text input field asking for the user's phone number, and a button handling submit. We will worry about the wiring later, as I'm currently working on fleshing out the backend functionality / running it on my local AWS account for alpha testing purposes. 

Down the line, this is where we can finally start managing some of our own backend :) 

## Motivation

After discussing with doctors, as well as product talent, we have decided that having a helpful, not bothersome or useless, notifications system would be a beneficial feature of the app. For the time being, this will send simple, helpful messages, but could be greatly expanded upon later and multiple services could be written out to handle multiple different types of notifications.

There is no alternative, users need to be able to send us their phone numbers on the frontend.

crispinamuriel:
fixes #224 

Added new feature of subscribe to alerts input field and submit button.
Button not hooked up, as issue did not require.

Future goal: have input hooked up to database so user can submit their phone number into 3rd party microservice for text phone alerts.

Future goal: have db signal this component's popup on condition if the user has not subscribed already


client/src/components/Subscribe.js
```diff
@@ -0,0 +1,56 @@
...
+            <TextField id="filled-phone" label="Phone Number" value={''} onChange={() => {}} variant="filled" />
+            <Button variant="contained" color="secondary">
+              Subscribe for alerts
+            </Button>
+          </FormControl>
+        </Slide>
+      </div>
+    );
+  }
+}
```
SomeMoosery: Try and make this a functional component, if possible

client/src/components/Subscribe.js
```diff
@@ -0,0 +1,56 @@
...
+  selected: {},
+  buttons: {
+    fontSize: '40px',
+
+    transition: 'all .3s ease-in-out',
+    '&:hover': {
+      transform: 'scale(1.1)',
+    },
+  },
+};
```
SomeMoosery: Frankly, I haven't been looking to see if other components have been following this, but we should extract all CSS into a related `Subscribe.css` file instead of inline with the JS

client/src/components/App.js
```diff
@@ -89,6 +90,7 @@ class App extends Component {
               <PrivateRoute path="/settings" authed={authed} />
             </Switch>
           </Layout>
+          <Subscribe />
```
SomeMoosery: We should put this in a component that doesn't persist across all components

client/src/components/Subscribe.js
crispinamuriel: Screen Shot 2020-03-26 at 11 59 52 PM

client/src/components/App.js
crispinamuriel: I need more information about when the user should see this alert in order to know where to place this component

client/src/components/Subscribe.js
crispinamuriel: I'll look into this

client/src/components/App.js
crispinamuriel: Okay last change made

PR #235, opened 2020-03-27
fixes #211 Disclaimer material ui styling

### LINKED ISSUE #211: Change disclaimer to Material-UI dialog
## Issue

As we are moving to be using Material-UI, we should restyle the Disclaimer to utilize the Material-UI library

## Motivation 

To make the transition to Material-UI and keep consistency in the app

Afrogrrrl:
Converted styling to material UI, deleted disclaimer.css as it was no longer necessary.


client/src/components/Disclaimer.js
```diff
@@ -1,37 +1,48 @@
...
 import actions from '../redux/actions/actions';
+import Button from '@material-ui/core/Button';
+import Dialog from '@material-ui/core/Dialog';
+import DialogActions from '@material-ui/core/DialogActions';
+import DialogContent from '@material-ui/core/DialogContent';
+import DialogContentText from '@material-ui/core/DialogContentText';
+import DialogTitle from '@material-ui/core/DialogTitle';
 
 export const Disclaimer = props => {
+  const [open] = React.useState(true);
```
SomeMoosery: Do we need this if it's never being changed? Or could we just set `open={true}` since it will always be open in the sense that we already determine if we display the dialog at all with the `{!props.answer ...` conditional

Afrogrrrl: I only did this cause it kept giving me an error when I didn't. I tried it and it works when open={true}. Pushed up the change.

SomeMoosery: Great!

Issue #187, opened 2020-03-24
Make new user onboarding component 


AdhamAH:
## Summary

Creat new user onboarding component as per [the wireframe](https://github.com/COVID-19-electronic-health-system/Corona-tracker/blob/master/design/wireframes/FINAL-V1-03.20/%233%20-%20New%20user%20onboarding.png)

## Motivation

This needs to be done as part of the plan for V2 as mentioned [in the product roadmap](https://docs.google.com/document/d/1W8QviIhwF9HDhCH4ww5dj97-ui6ouiVh6fGsNUCxj4I/edit#)

fallon7284:
Is anyone working on this? It's unclear, it's marked in-progress [in the product roadmap](https://docs.google.com/document/d/1W8QviIhwF9HDhCH4ww5dj97-ui6ouiVh6fGsNUCxj4I/edit#).  If nobody is working on it I'll take it.

SomeMoosery:
All you @fallon7284 - we should stick to GitHub Issues as the source of truth for what's being worked on

AdhamAH:
@fallon7284 thanks for working on it! Please try to keep in mind using Material-UI styling in doing this but use the jsx normal input tag.

fallon7284:
Well thats going to take some reworking, it's nearly finished but I've done it all with Material-UI components

PR #239, opened 2020-03-27
#187 OnboardUser.js new user onboarding component

fallon7284:
Description of change => Adds a user onboarding component, based off wireframe.  Simple react form, component holds state locally, component can receive a function as a prop I called postNewUser which would be the function to post the user to the database.  That function *when written and passed to this component once it is connected to redux* should post the new user to the database and then receive the response and persist it in redux.

Possible drawbacks => CSS needs some work, little issue fixing the with the fonts on the sex buttons, and the layout isn't perfect (I was working on this before the logos and the navbar were moved to top level, so I sort of hastily removed them before this commit, which affected layout a bit).  Could use some validation on age field (is Number(age) an integer?) and maybe on location data.  Doesn't currently even validate that these fields are non-empty, so could be considered a work in progress, but it seemed ready to PR.


client/src/components/OnboardUser.js
```diff
@@ -0,0 +1,233 @@
...
+const blankForm = {
+  firstName: '',
+  age: '',
+  sex: '',
+  city: '',
+  state: '',
+  zip: '',
+};
+
+export default ({ postNewUser }) => {
```
SomeMoosery: Doesn't this component need a name?

client/src/components/OnboardUser.js
```diff
@@ -0,0 +1,233 @@
...
+  const [formState, setFormState] = useState(blankForm);
+  const handleChange = e => {
+    e.preventDefault();
+    setFormState({
+      ...formState,
+      [e.target.name]: e.target.value,
+    });
+  };
```
SomeMoosery: I think we should move this state logic to Redux, right? Since we'll potentially need this onboarding information elsewhere in the app (when the user opens the settings page, for example).

client/src/components/OnboardUser.js
```diff
@@ -0,0 +1,233 @@
...
+  const classes = useStyles();
+  return (
+    <div className={classes.root}>
+      <div className={classes.header}>
+        <h4>
+          <b>Welcome to CoronaTracker!</b>
+        </h4>
```
SomeMoosery: I think we should be defining HTML elements using Material-UI styling, so a `Typography variant='h4'` in this case

client/src/components/OnboardUser.js
fallon7284: I can give it a name within the file if you like.  It's a default export so the name in the file doesn't affect the name when you import it.

client/src/components/OnboardUser.js
fallon7284: You wouldn't want to use redux to hold the state of the fields in a form before it is submitted.  I noted in my PR that the flow would ideally be, this component receives a function that, when the user clicks "SAVE MY RESPONSES", posts the form data to a database, and then posts the response of the created data to redux.  You'd want to persist data to redux only after you know that it has been persisted in the database in order to keep your application in sync.  The useState you see here only holds the data while this form component is open.

client/src/components/OnboardUser.js
fallon7284: I've got to be honest, this component was basically finished last night and I was pretty frustrated when @AdhamAH noted

"@fallon7284 thanks for working on it! Please try to keep in mind using Material-UI styling in doing this but use the jsx normal input tag."

because I had already done it in Material-UI components and those guidelines had not been noted in the issue.  I went back today and changed it all to jsx.  This feels like a nitpick, no?

client/src/components/OnboardUser.js
SomeMoosery: Ah okay, you're right. Sorry, I shouldn't be going through these PRs at work - I'm not reading carefully enough.

client/src/components/OnboardUser.js
SomeMoosery: Cool, didn't realize that was a feature. I'll leave it up to you if you want to add one or not, but it may just be useful for consistency's sake then.

fallon7284: Fixed.

client/src/components/OnboardUser.js
fallon7284: No sweat.

fallon7284:
Screen Shot 2020-03-27 at 1 10 34 PM

client/src/components/OnboardUser.js
SomeMoosery: Okay, let's leave this for now. This is nitpicky and definitely doubles down on the need for an open, easily-accessible style guide. I'll reach out to you to see what action we can take regarding this.

AdhamAH:
fixes #187

fallon7284:
Yes

AdhamAH:
thanks, fantastic work by the way!

Issue #223, opened 2020-03-26
Survey Screen is not Centered on Desktop


SomeMoosery:
### Prerequisites

* [ X ] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in QA
    * Followed all applicable steps in the TROUBLESHOOTING guide: TODO
    * Checked for relevant discussions on Discord: <https://discord.gg/pPERUuv>
    * Checked that your issue isn't already filed: <https://github.com/COVID-19-electronic-health-system/Corona-tracker/issues>

### Description

On desktop, the survey displays skewed off to the right, with the title and sign out buttons displaying below it.

<img width="1388" alt="Screen Shot 2020-03-26 at 7 53 16 PM" src="https://user-images.githubusercontent.com/15767602/77707581-ea50cf80-6f9b-11ea-87b5-2776de9468dd.png">

### Steps to Reproduce

1. Log in on deskotp
2. Take survey

**Expected behavior:**

The survey appears stylistically in the center of the page, with the "constant" features (logo, sign out) above it

**Actual behavior:**

The survey is offset far off the the right side of the screen, and the logos, signout button, etc are positioned below it.

**Reproduces how often:**

100%

crispinamuriel:
May I be added to this issue? (I would welcome anyone else to try and fix it as well!)

SomeMoosery:
All you 👍

PR #243, opened 2020-03-27
fixes: #223 survey is now centered

crispinamuriel:
Fix: Survey page is now centered


crispinamuriel:
Screen Shot 2020-03-26 at 11 50 25 PM

crispinamuriel:
Screen Shot 2020-03-27 at 2 56 45 PM

crispinamuriel:
I've rebased, Thanks Carter.

SomeMoosery:
Hm, it looks like there's still old files being changed (for example, in master now we're using `authenticated` not `authed` in `App.js`... did you pull in the most recent changes from upstream/master before rebasing? I'm not sure if that's exactly the fix, but might point you in the right direction. Apologies, I don't have too much experience working with git on forked copies.

Issue #204, opened 2020-03-26
Split components/.../CheckboxButton.js into multiple files


fallon7284:
## Summary

components/survey-view/checkbox-button/CheckboxButton.js exports a react component, but also contains Redux action types, action creators, initial state, and reducers.

Should try to maintain a consistent file structure by moving the Redux code into the src/Redux/actions and src/Redux/reducers.

## Motivation

Maintain consistent file structure

Afrogrrrl:
Would be happy to take this one on!

liadove:
for my understanding useReducer is not the same as Redux. useReducer helps to control inputs in this case. If I understand it right?

SomeMoosery:
Yeah, `userReducer` is separate from Redux, but since for the time being we've decided to move forward with Redux, this story would be more of a refactoring of this code to utilize Redux, right?

SomeMoosery:
And sounds good @Afrogrrrl !

fallon7284:
Absolutely right, I completely misunderstood what was happening in this file.  This is all being done just to handle form inputs within this component.  Mea Culpa.

SomeMoosery:
No problem! It's a great point-out though, this may have gone unnoticed not being in Redux for some time, and we want to rely on it for anything non-trivial 👍 . May be worthwhile to edit the issue to more contextualize refactoring this to use Redux

fallon7284:
I wouldn't even think this gets refactored to Redux, though, now that I understand what it is.  It's holding form state pre-submission.  I don't think you want to use the redux store to hold form state within a single component. 
I think through my misunderstanding I unfortunately created work that didn't need to be done.  I'm not super comfortable editing the issue and would just hope to bring @Afrogrrrl into the conversation since she's working on this now ill-defined task.
Apologies for the mix up.

liadove:
@fallon7284 That's totally fine! I wasn't clear when I wrote it, maybe I should've used regular useState instead? But I thought we can add more symptoms in it, and it would be easier to put it in useReducer. And also it much less code!

SomeMoosery:
Ah, okay, I see what's going on now. Apologies about that. You're right, this is trivial in the sense that we shouldn't be storing like intra-component state in Redux. I think `useReducer` actually maybe was best since it is more likely that `useReducer` is more reusable. 

@Afrogrrrl - would you be willing to take on another issue instead of refactoring this? Apologies for the confusion I caused.

[...]

PR #246, opened 2020-03-27
Redux for symptom tracker survey

liadove:
fixes https://github.com/COVID-19-electronic-health-system/Corona-tracker/issues/204
I thought you were right we need to save all responses before they submit the survey, in case if they press on other tab and want to come back to survey


SomeMoosery:
Needs a rebase off of master. I'm sorry, but I'm honestly still a little bit confused why we feel it necessary to store state in redux. Aren't we just maintaining this user state within the component? And don't necessarily need to store anything to Redux? I think it's fair to say that if the user navigates away from this page their progress can be lost, as we only want to save responses when they submit the survey, otherwise things can get messy/confusing.

@fallon7284 I know you explained this in #204 which is why we closed it. Am I missing anything?

fallon7284:
I'm running out for a while and won't be able to give this PR a real look until a good bit later...
Having said that, the flow I'd expect from a survey submission, regarding like local state, database, redux would be: Form fields are stored in state inside the survey component.  Connected component gets a thunk action creator from redux.  Thunk handles asynchronously posting the survey data to the database and then persisting the data from the response in the redux store.  I don't 100% understand the desired flow of the app or what data we expect to use in the front end, @SomeMoosery, so I don't totally understand whether you'd want that survey data added to the redux store at all, though my gut says that you would, so that you could, for example, have components that show the change in survey answers over time.  But certainly you'd only want to persist the data to the redux store as a result of first posting it to the database.

SomeMoosery:
Sounds good, and no problem! Thanks for the good writeup. Yeah, that's what I was really getting at @fallon7284 - I don't think we need to be storing state of the component as it's being filled out in Redux, only on submission. As we chatted about, I'll get a much clearer vision of application flow / what we need at each component shortly.

liadove:
We can store a state of the component in the component itself, this is what I was thinking too. But would it be helpful for users if they clicked on another tab by mistake have everything they already filled out?

SomeMoosery:
I think this should rely more on data caching or user sessions than on Redux, though

PR #258, opened 2020-03-28
HOTFIX: Changes necessary to deploy React build

SomeMoosery:
No issue as these are necessary fixes that needed to be completed on the spot for the build to deploy to S3.

### DIFF
--- a/client/public/index.html
+++ b/client/public/index.html
@@ -7,7 +7,7 @@
     <meta name="theme-color" content="#000000" />
     <meta
       name="description"
-      content="Web site created using create-react-app"
+      content="Corona Tracker"
     />
     <link rel="apple-touch-icon" href="%PUBLIC_URL%/logo192.png" />
     <!--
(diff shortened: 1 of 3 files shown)


tesla809:
LGTM!

tesla809:
Type error causing CI tests to fail.

Error messages pop up:

Error: Uncaught [TypeError: Cannot read property 'dataSample' of undefined]

  TypeError: Cannot read property 'dataSample' of undefined
    >  8 | const dataSample = useSelector(state => state.root.dataSample);

SomeMoosery:
We can leave this open until it's resolved, but I think this is an error with our current package management. Even a fresh install of master fails these tests now.

I think we should be more diligent about ensuring everyone is using yarn, and that every PR is rebased off the most current master. Babel 7.8.7 has been sneaking back in constantly, and will break tests/builds as it was a broken release

tesla809:
Great points @SomeMoosery. Working on the CONTRIBUTING.md and adding the language around rebasing off the current master. 

I will add the information to alert the devs only to use Yarn.

SomeMoosery:
Actually, while it's good we stay up to date in our package.json of course, this also looks to be a more systemic problem. It's `create-react-app`'s [pinned issue](https://github.com/facebook/create-react-app/issues/8680) right now

PR #261, opened 2020-03-28
feature: added map component

Ahwan0717:
Closed my previous PR so I could make a new one. Map component was created using mapbox-gl and is scrollable. I'm working on getting Travis to pass.
Screen Shot 2020-03-28 at 12 55 33 AM


client/src/components/Map.js
```diff
@@ -0,0 +1,55 @@
...
+        if (!map) initializeMap({ setMap, mapContainer });
+    }, [map]);
+
+
+    return (
+        <div className="mapContainer">
+            <h4>Check out the Map!</h4>
+            <h4>...COMING SOON...</h4>
+            <div ref={el => (mapContainer.current = el)} style={styles} />
+            <NavBar />
```
SomeMoosery: We shouldn't be adding `NavBar` anywhere else than it is now - it should be a top-layer component in `Layout`

client/src/components/Map.js
```diff
@@ -0,0 +1,55 @@
...
+import '../css/Map.css'
+
+const mapboxToken = process.env.REACT_APP_MAPBOX_TOKEN
+
+
+const styles = {
+    width: "100vw",
+    height: "calc(100vh - 80px)",
+    position: "absolute"
+};
```
SomeMoosery: We should be using Material-UI `withStyles`, I believe

client/src/css/Map.css
```diff
@@ -0,0 +1,5 @@
+.mapContainer {
+    padding-top: 3vh;
+    text-align: center;
+    height: 100vh;
+}
```
SomeMoosery: We should keep all CSS for a given component in the same place, ideally in `withStyles`

client/src/components/Map.js
Ahwan0717: Got it, will adjust accordingly tomorrow (later today). Thank you for all of your feedback @SomeMoosery.

AdhamAH:
@SomeMoosery might be busy. I think it is good to go. Thanks

PR #263, opened 2020-03-28
Use blockstack hook -- closes #244 

### LINKED ISSUE #244: Use react hook for authentication status and Blockstack session
### Description

With #216 the authentication status and other functionality from the Blockstack SDK is available in function components through the `useBlockstack()` hook from the [react-blockstack](https://github.com/njordhov/react-blockstack) library, as in:

```javascript
import { useBlockstack } from 'react-blockstack';

function SignOutButton (props) {
  const { authenticated, userSession, signOut } = useBlockstack();
  return ( <button disabled={!authenticated} onClick={signOut}>SignOut</button>)
}
```

This should supersede passing down authentication status through the `authed={authenticated}` prop.

fallon7284:
fixes #244 

**Description of change** => Uses useBlockstack hook inside components that need access to userSession object, authenticated boolean, signIn/signOut methods which we previously passed as props.

**Alternate design** => Continue to pass as props -- would be bad for readability and ease of development.

**Possible drawbacks** => Making everyone working on this project aware of the hook and what it provides will be necessary, otherwise it will be hard to keep the codebase consistent.  But this approach, if implemented consistently, lends itself to much cleaner code.


SomeMoosery:
LGTM

PR #271, opened 2020-03-29
Store user data in Gaia

### LINKED ISSUE #237: Persist Survey Results to Database
## Summary

- [ ] On completing a survey, an Observation is created and persisted to the database using Radiks.

## Motivation

We need a way to save survey responses, and should start persisting this to the database.

SomeMoosery:
fixes #237 

## Changes
- Added capability to save and fetch data from Gaia on survey submit 
- Store data as a new file each time so that we can iterate through all user data.

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -1,6 +1,6 @@
 {
   "name": "corona-tracker",
-  "version": "0.1.0",
+  "version": "0.2.0",
   "private": true,
   "prettier": "@blockstack/prettier-config",
   "dependencies": {
--- a/client/src/components/SymptomsTracker.js
+++ b/client/src/components/SymptomsTracker.js
@@ -1,7 +1,8 @@
-import React, { useState, useRef } from 'react';
+import React, { useState, useEffect, useRef } from 'react';
 import { makeStyles } from '@material-ui/core/styles';
 import { Typography, Slider, ButtonGroup, Button, TextField, Grid } from '@material-ui/core';
 import CheckboxButton from './survey-view/checkbox-button/CheckboxButton';
+import { useBlockstack } from 'react-blockstack';
 
 // custome style for material ui elements
 const useStyles = makeStyles(theme => ({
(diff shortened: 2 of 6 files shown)


njordhov:
Do you by chance use react-blockstack hooks with Redux? It would explain the useAtom mention (which is used by react-blockstack). Hooks can only be used within function components.

njordhov:
``` 
decrypt: true
```
The Blockstack SDK encrypts by default so no need to explicitly enable encryption or decryption.

SomeMoosery:
Sounds good, thanks! I'll make some adjustments and add a way to incrementally add observations in the morning.

njordhov:
Suggest you keep each observation in separate read-only files, and either keep an index file or build the list of observations using `listFiles()`. This is a bit primitive, but will be resilient as you can restore the observations on screw-ups and not risk corrupting stored observations if you accidentally introduce a bug in storage.

njordhov:
See the documentation for [listFiles()](https://blockstack.github.io/blockstack.js/globals.html#listfiles) as it appears the way you're using it won't work as expected.

SomeMoosery:
Made some updates, this hopefully allows automatic incrementation of observations on each submission. Thoughts?

SomeMoosery:
I think this should be good to go once tests pass. We're able to incrementally add surveys to Gaia and then fetch them from the database where needed. Now, we can start populating other components with real user data.

njordhov:
Suggest you use a directory prefix separated with `/` for the observations, like in `observation/${numObservations+1}.json` to standardize on the conventional pathname scheme. Later you can get the keys for observations by filtering the list of filenames on the directory path. Also, zero-pad the filename as in `00000001.json` to a preset number of digits so they sort alphabetically also when you go beyond single digit.

SomeMoosery:
Very smart. I'll open up a new PR in just a little bit adding this in

PR #275, opened 2020-03-29
Finished investigating @shopify/react-i18n- for handling translations

### LINKED ISSUE #118: Investigate: @shopify/react-i18n- for handling translations, formatting, and more.
## Summary
Investigate [@shopify/react-i18n](https://www.npmjs.com/package/@shopify/react-i18n) it is a i18n utilities for React handling translations, formatting, and more.

Architecture wise for i18n, code using properties (I.e key-value properties file).

This can be used for our translation into different languages.

See:
[https://en.wikipedia.org/wiki/Internationalization_and_localization](https://en.wikipedia.org/wiki/Internationalization_and_localization)

## Motivation
**Why are we doing this?** 
We need a way to use our translations in our file. This package helps with this.
We want our app to inform people in other countries before the wave of community infections start to overwhelm their health systems, especially in places with poor infrastructure. 

We can leverage the fact that internet connectivity is widespread to educate and share information.

**What use cases does it support?** 
Supports translation team's efforts

**What is the expected outcome?**
We have an already built way to support our translations in the file.

Any alternatives can be considered. This was recommended by the systems architect advisor who has experience with this.

akilhylton:
fixes #118


akilhylton:
[Imgur](https://i.imgur.com/rhRepW3.png)

SomeMoosery:
This all looks like awesome work to me, but would you just be able to provide a bit of a write-up on the things changing / being done by this PR? This seems like a huge step but I just want to be sure there's clarity.

Also, would you mind just adding "fixes #<ISSUE>" of whichever issue this references in the body of the PR? 

Seriously, though - this looks like a major addition, super exciting stuff 🚀

PR #284, opened 2020-03-29
#252 adjust symptom tracker layout based on size of screen

### LINKED ISSUE #252: Large Screens view
## Idea

To make the view in large screens inline like this:

Screenshot 2020-03-27 at 22 22 13
**just an example, not end result**

## Motivation 

To make the use of the app easier to read and follow in a big screen

Avidness:
fixes #252 

covid-responsive-ui


SomeMoosery:
Looks great, thanks @Avidness

PR #299, opened 2020-03-30
Persist disclaimer answer

### LINKED ISSUE #283: Persist Disclaimer Answer to Gaia
## Summary

While we should definitely continue to store user's disclaimer agreement in Redux for easier access when rendering components, we should also persist this to the database so that the next time the user logs in, they won't have to re-consent to the agreement.

A long-term change to this could be, on login, doing a one-time database read to see if the user has already consented. If not, the disclaimer is displayed and acts as it does now. If so, we will then store that the user has consented in Redux for all future traversal around the app during that user session.

We will probably just want to upload a file `"disclaimerConsent.json"` with one or two attributes, `consentStatus` and `date`.

**Please continuously check [this PR](https://github.com/COVID-19-electronic-health-system/Corona-tracker/pull/271) relating to our "first contact" with Gaia to ensure we're following the same format and staying up to date.**

## Motivation

If the user is greeted with a disclaimer every time they log into the app, they may become frustrated. If we can alleviate a little it of that frustration by not showing the disclaimer on login each time, users will be more likely to return.

We keep the login functionality as it is now

Afrogrrrl:
writes disclaimer answer to Gaia. fixes #283


client/src/components/Disclaimer.js
```diff
@@ -9,8 +9,26 @@ import DialogActions from '@material-ui/core/DialogActions';
...
+  const storeAnswer = answer => {
+    props.setAnswer(answer);
+    disclaimerAnswer.answerChoice = answer;
+
+    try {
+      userSession.putFile(`disclaimer.json`, JSON.stringify(disclaimerAnswer));
+    } catch (error) {
+      console.error(error);
+    }
+  };
```
SomeMoosery: We should try to keep these functions consistently asynchronous across the board as outlined by Blockstack

secrets.env
```diff
@@ -0,0 +1,4 @@
+REACT_APP_QA_URL=http://ec2-54-92-173-146.compute-1.amazonaws.com:1260
+export MONGODB_URI="mongodb://admin:foobar1@localhost:27017/test1"
+
+//blockstack key: youth already cream trade core frequent lumber quit engine recall chief divorce
```
SomeMoosery: Please remove this

Afrogrrrl: whoops, don't know how that made it in there

client/src/components/Disclaimer.js
SomeMoosery: Sorry to be pedantic, but I was thinking we keep things consistent with the Blockstack documentation by doing something like 

```
await userSession.putFile(...).then(res => {
    doSomething(res)
}).catch(err => 
    doSomething(err)
});
```

Just so we keep our styles consistent across components. If you have a reason why that's frivolous though definitely let me know!

Afrogrrrl:
sorry, not sure why package-lock.json keeps getting pushed up, it wasn't happening before, I'm not sure what changed. but I added in the await so hopefully it's good to go now

njordhov:
The `useFile()` hook from [react-blockstack](https://github.com/njordhov/react-blockstack#persistent-data-storage) can be used for data persistency in components.

    import { useFile } from 'react-blockstack'

    const [content, setContent] = useFile("disclaimer.json")

client/src/components/Disclaimer.js
njordhov: I recommend the `useFile()` hook from react-blockstack in function components, rather than accessing the blockstack file api directly. It covers the functionality of putFile/getFile/DeleteFile but is reactive.

SomeMoosery:
Ah, that makes sense! I'll write a ticket to refactor these to implement `useFile()` 👍

Afrogrrrl:
That probably won’t be necessary because all of this is about to into a thunk. I’ll refactor it into useFile() there.

SomeMoosery:
Since I already wrote an issue for the refactor, when you submit a PR if you could just close both the thunk issue and the refactor issue that would be great 👍

Issue #281, opened 2020-03-29
Multiple Warnings/Errors on Survey Load


SomeMoosery:
### Prerequisites

* [ X ] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in QA
    * Followed all applicable steps in the TROUBLESHOOTING guide: TODO
    * Checked for relevant discussions on Discord: <https://discord.gg/pPERUuv>
    * Checked that your issue isn't already filed: <https://github.com/COVID-19-electronic-health-system/Corona-tracker/issues>

### Description

When opening the survey, there are multiple errors relating to DOM elements that should be resolved to ensure our code works smoothly and as planned 100% of the time.

<img width="1380" alt="Screen Shot 2020-03-28 at 2 13 09 PM" src="https://user-images.githubusercontent.com/15767602/77854236-0b413c80-71b7-11ea-92bd-d68fcdb63690.png">

### Steps to Reproduce

1. Log in
2. Navigate to survey and inspect

**Expected behavior:**

No errors/warnings

**Actual behavior:**

Errors/warnings

**Reproduces how often:**

100%

JWaltuch:
I can take this :)

JWaltuch:
I figured out the main errors and one of the warnings.

I'm still getting this warning:
./src/components/Login.js
  Line 17:9:  'changeLanguage' is assigned a value but never used  no-unused-vars
But I'm not certain how/where change language is supposed to be used?

JWaltuch:
Oh cool. Then I'll make a pull request because I handled the error.

SomeMoosery:
Okay nice, just be sure to add the error log from chrome inspector in your PR. Maybe that error you posted inadvertently has to do with survey, I'm not sure, I'll need some autonomy on this one to just make sure all "red" errors are fixed.

JWaltuch:
No problem :)

SomeMoosery:
Reopening: unfortunately, previous changes broke the UI when displayed on mobile, so we should find a different implementation of how to fix errors

AdhamAH:
@JWaltuch if you having trouble fixing this, let me know I can take it

SomeMoosery:
Closing, fixed by #303

PR #303, opened 2020-03-30
Changed Grid to ButtonGroup, Issue 281

JWaltuch:
image
image
image

I changed Grid to ButtonGroup to fix the error. fixes issue #281 

Is this more what you're looking for @SomeMoosery? I'm not sure if this is the style you wanted, but it appears to look the same as the currently deployed version, but there is no error.


SomeMoosery:
Could you just test this with the responsive viewport in Chrome Dev Tools as well to make sure the layout looks the same?

https://developers.google.com/web/tools/chrome-devtools/device-mode

JWaltuch:
Tested with responsive. Seems to have the same behavior as in the current deployed version so that should be all good :) Also bumped up the version

PR #314, opened 2020-03-31
refactored disclaimer gaia write into thunk

### LINKED ISSUE #298: refactor all Gaia read/write requests into Redux thunks
## Summary

As we start hooking up components to read/write to Gaia we should take advantage of Redux and put these read/write requests into thunks. 

For those unfamililar with thunks, you can read more about them [here](https://github.com/reduxjs/redux-thunk). They are essentially a form of middleware. They perform calls to our dB asynchronously and update the store. . 

## Motivation
This is good practice because should errors occur reading/writing to Gaia it will protect stability of the app, since the error will be "caught" in the thunk and not passed along to live components. It will also make it easier to debug should these errors occur.

alternative is to keep making direct calls to Gaia within components

## Additional context

related to issue #256

Afrogrrrl:
fixes #298 

refactored gaia call in disclaimer component into redux thunk


SomeMoosery:
LGTM!

PR #321, opened 2020-03-31
Disclaimer will not load if user has already agreed

### LINKED ISSUE #319: Don't show disclaimer if user has already consented
## Summary

When the user logs in, use `useFile` ( #306 ) to get `"disclaimer.json"`, if it exists. If so, and it indicates that the user has already consented, do not display the disclaimer.

## Motivation

We want to make the UX as seamless as possible. If we can rid the user of a popup each time s/he logs in, that fraction of a second amounts to a much more pleasant UX.

Leaving as is

JWaltuch:
fixes #319 
image
image

Used useFile hook to check for "disclaimer.json". If found, will not load disclaimer component on page load.


SomeMoosery:
LGTM!

Issue #351, opened 2020-04-01
Temperature data not saved in the backend


pavel-ilin:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in QA
    * Followed all applicable steps in the TROUBLESHOOTING guide: TODO
    * Checked for relevant discussions on Discord: <https://discord.gg/pPERUuv>
    * Checked that your issue isn't already filed: <https://github.com/COVID-19-electronic-health-system/Corona-tracker/issues>

### Description

Related to issue #333 and #339 

When I open single observation json file from the Blockstack in DiagnosticContainer there is no temperature data in it. There is nothing to extract and pass to the Chart component.

### Steps to Reproduce

1. Open DiagnosticContainer
2. Add line   "const [observation] = useFile("observation/0000001.json");"
3. console.log(observation)

**Expected behavior:**

observation/0000001.json should contain numeric value which represents temperature which user submitted in SymptomsTracker component. 

**Actual behavior:**

observation/0000001.json doesn't contains numeric value which represents temperature which user submitted in SymptomsTracker component.

**Reproduces how often:**

100%

SomeMoosery:
Is anything being saved in the database - like, is temperature being stored as a string? Or is nothing being saved at all?

If the former, we should just convert it to a number before posting to gaia - if the latter... that might be more of a problem with that specific question. 

Good catch! Are you willing to tackle this as well? Or should we leave this open for another contributor to take on while you work on the charts?

pavel-ilin:
It think it's a problem with specific question. 
That's what I Blockstack give back:
![response](https://i.ibb.co/vk25LXF/Screen-Shot-2020-04-01-at-3-52-44-PM.png)

We can see that we have boolean value for the fever, but there is no key and value which will show us how high fever is.

Also I having a problem with using useFile hook for multiple files. If someone has experience work with it I would be happy to learn.

SomeMoosery:
Hm... yeah, I'm not sure why off the top of my head but it looks like we aren't capturing [fever severity](https://github.com/COVID-19-electronic-health-system/Corona-tracker/blob/37214e8d9b9b5857149d9edb2114da68285c6db0/client/src/components/survey-view/checkbox-button/CheckboxButton.jsx#L108) for some reason. 

[...]

PR #352, opened 2020-04-01
temperature

liadove:
fixes https://github.com/COVID-19-electronic-health-system/Corona-tracker/issues/351

The bug was caused by not correctly handling the event in TemperatureAutocomplete subcomponent.


SomeMoosery:
Nice catch!

PR #364, opened 2020-04-02
Redirect to onboard screen on disclaimer agree (#330)

### LINKED ISSUE #330: Redirect to Onboard User if no disclaimer found
## Summary

When logging in, as we're already checking to see if there exists a `"disclaimer.json"` in Gaia storage for the associated user in `DiagnosticContainer`, we should use that to also determine whether or not to redirect users to an initial onboarding screen (`OnboardUser`). We would probably need to utilize the `disclaimer` action/reducer to add a field for whether or not this is the user's first time on the application (i.e. if `disclaimer` was null, it's the user's first time), and then show the onboarding after the user consents to the disclaimer.

- [ ] if `disclaimer === null`, set new redux state variable that this is the user's first login
- [x] after disclaimer completion, if the user has not navigated away, display the user onboarding screen

## Motivation

We want the UX to be as smooth as possible, and prompting the user for a one-time signup will help with that

Just keep the user onboarding separate and manual

acthelemann:

Fixes #330

After agreeing to the disclaimer, the user will now be redirected to the onboarding screen. This creates a more appropriate flow that encourages the user to input their information as soon as possible.

Previously, the disclaimer modal would only appear on the Log screen if the user had not seen it before, but not necessarily agreed to it. Now, the disclaimer modal appears on any screen after logging in, and will only disappear after the user agrees to it. This will help ensure that the user agrees to the disclaimer before using any part of the app.

### Test plan (required)

To test locally:
1. Create a new account, or use an account that has not agreed or disagreed to the disclaimer.
2. Navigate to any page by typing the URL manually (http://localhost:3000/, http://localhost:3000/education, etc.). The disclaimer modal should appear.
3. Click the disagree button for the disclaimer. The disclaimer modal should not disappear.
4. Refresh the page. The disclaimer modal should appear.
5. Click the agree button for the disclaimer. The disclaimer modal should disappear and the onboarding screen should be navigated to automatically.
6. Refresh the page now. The disclaimer modal should not appear.

### Final Checklist

    - *I feel second decimal should be for minor change, and third for patch/bug fix, but I followed the recommendation below.*

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -1,6 +1,6 @@
 {
   "name": "corona-tracker",
-  "version": "0.2.9",
+  "version": "0.2.10",
   "private": true,
   "prettier": "@blockstack/prettier-config",
   "husky": {
--- a/client/src/components/App.jsx
+++ b/client/src/components/App.jsx
@@ -4,7 +4,8 @@ import { configure, User } from 'radiks';
 import { Connect } from '@blockstack/connect';
 import { BrowserRouter, Switch } from 'react-router-dom';
 import { connect } from 'react-redux';
-import ReactBlockstack, { useBlockstack, didConnect } from 'react-blockstack';
+import ReactBlockstack, { useBlockstack, didConnect, useFile } from 'react-blockstack';
+import Container from '@material-ui/core/Container';
 import Layout from './Layout';
 import Map from './Map';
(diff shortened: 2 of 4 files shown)


SomeMoosery:
🚀 🚀

PR #366, opened 2020-04-03
survey displays if has not been submitted

### LINKED ISSUE #215: Redirect to Symptom Tracker Survey when user login
## Summary

Add functionality to display survey on login, if survey hasn't yet been completed for the day

## Motivation

We want people to submit the survey every day, and visualizing it when they didn't submit it yet will help us remind people to do so.

NGabuaeva:

fixes #215 

If the survey hasn't been submitted for the day, after logging in survey page will be rendered, otherwise the home page will be displayed. I am using window.localStorage to solve this problem, can add the code for it for browsers that don't support localStorage if necessary. 
### Test plan (required)

Ran the app locally two days in a row, if the survey has not been submitted the first page the user will see after logging in will be the survey. If the user logs out and logs back in the same day the home page will be displayed.

### Final Checklist

- [ ] Have you bumped the version in `package.json`?
- [ ] Is your PR rebased off the most current master?
- [ ] Have you squashed all commits, or if you will merge will you squash all commits?
- [ ] Did you use yarn, not npm?
- [ ] Did you use Material-UI wherever possible?
- [ ] Did you format according to Prettier?
- [ ] Did you run all of your most recent changes locally to make sure everything is working?


client/src/components/DiagnosticContainer.jsx
```diff
@@ -23,31 +24,49 @@ const dateOptions = {
   day: 'numeric',
 };
 
+// check if the survey has been submitted today
+const hasSubmitted = () => {
+  const date = window.localStorage.getItem('date');
+  const todaysDate = new Date().toISOString().slice(0, 10);
+  if (date === todaysDate) {
```
SomeMoosery, line 31: Would there be a case where these two aren't equal?

yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
+# yarn lockfile v1
+
+
```
SomeMoosery, line 4: Please delete this `yarn.lock` in the root directory

client/src/components/DiagnosticContainer.jsx
```diff
@@ -23,31 +24,49 @@ const dateOptions = {
...
+    if (window.localStorage.getItem('surveyCompleted') === 'false') {
+      return false;
+    }
+    return true;
+  }
+  window.localStorage.setItem('date', todaysDate);
+  window.localStorage.setItem('surveyCompleted', 'false');
```
SomeMoosery, line 38: Is there any benefit of using local storage here and not storing the surveyCompleted state in the Redux store?

CONTRIBUTING.md
```diff
@@ -378,33 +378,33 @@ Unsure where to begin contributing to Atom? You can start by looking through the
...
 
-See the full list of commands [here](https://classic.yarnpkg.com/en/docs/cli/).  
+See the full list of commands [here](https://classic.yarnpkg.com/en/docs/cli/).
 
-`yarn install`  
-Installs all node modules  
-Similar to `npm i`  
+`yarn install`
+Installs all node modules
+Similar to `npm i`
```
SomeMoosery, line 385: These changes shouldn't be here as it adds unnecessary file changes to the PR

client/src/components/DiagnosticContainer.jsx
NGabuaeva: Yes. If I am logging in today today's date will be stored in the local storage, so when I login again tomorrow they will not be equals, so the function will return false and I will be shown the survey after logging in

client/src/components/DiagnosticContainer.jsx
NGabuaeva: My thinking was that the localStorage is more persistent, so if the user opened the app in the morning, filled out the survey, then logged out, closed the app, later when the user opens the app again the same day, the survey won't be displayed after logging in. This way we can make sure that the user gets the survey as the first page only once a day

NGabuaeva:
Before login
After login for the first time today 2020-04-03 at 12 52 42 PM
After logging in again the same day 2020-04-03 at 12 53 56 PM

Let me know if these work

yarn.lock
NGabuaeva: Deleted in the last commit

CONTRIBUTING.md
NGabuaeva: I haven't touches this file, not sure why it even showed up in changes, but should be fixed in the last commit

NGabuaeva:
Could you tell me about the rebase off of master? I pulled from upstream, is there more I need to do?

client/src/components/DiagnosticContainer.jsx
SomeMoosery: Makes sense 👍

SomeMoosery:
All changes looking good me now, thanks for the explanations! It looks like you've rebased properly as well - there's some guides in our CONTRIBUTING.md around it for future reference but it looks like you did it properly here.

One more thing I forgot to mention and then this is good to go - if you could just bump the `version` in `package.json` in the client, that would be great!

NGabuaeva:
Bumped the version from 0.2.10 to 0.2.11, is that correct?

SomeMoosery:
Yup, exactly!

PR #388, opened 2020-04-05
Uses real chart data and fixes styling

### LINKED ISSUE #387: Feed in user data to charts
## Summary

- [ ] Bar chart gets real user data for all observations
- [ ] Line chart gets real user data for all observations

## Motivation

For beta, we need all of our components to be interacting with real user data, not dummy data.

There aren't really any, need that data

SomeMoosery:

fixes #387 

- Updates styling for charts
- Passes in user data for dates, temperatures
- **Using this map, we could pretty easily pass in a different background color (for example) if the user's temperature was over a specific amount in the future if we want**

charts

### Test plan (required)

- Not much new to add here, can add E2E tests post-beta

### Final Checklist

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -1,6 +1,6 @@
 {
   "name": "corona-tracker",
-  "version": "0.2.21",
+  "version": "0.2.22",
   "private": true,
   "prettier": "@blockstack/prettier-config",
   "husky": {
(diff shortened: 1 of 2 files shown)


pavel-ilin:
For some reasons entered data is not appeared on the Chart.

Data being entered:
entered data

Empty Chart:
empty chart

SomeMoosery:
Hm, and there's 4/5 on the x-axis right?

I'd file a bug report, since (famous last words) it's working on my machine

Issue #469, opened 2020-04-11
Button Functionality in "More" Tab


salvolpe:
If I click on `More`, I have to click specifically on the text in order to advance myself to the screen of interest. This applies ONLY to `Settings` and `About`. It works perfectly fine for `Sign Out`, i.e. I can click anywhere on that button and it will sign me out.

[![Image from Gyazo](https://i.gyazo.com/8ab943d62c0acde7ff152be3b4b0a82a.gif)](https://gyazo.com/8ab943d62c0acde7ff152be3b4b0a82a)

AlexRAV:
Hi! I can take it.

PR #471, opened 2020-04-11
Fix buttons behavior in More tab

AlexRAV:
fixes https://github.com/COVID-19-electronic-health-system/Corona-tracker/issues/469

fix2

### DIFF
--- a/client/src/components/More.jsx
+++ b/client/src/components/More.jsx
@@ -6,7 +6,7 @@ import Button from '@material-ui/core/Button';
 import Dialog from '@material-ui/core/Dialog';
 import DialogActions from '@material-ui/core/DialogActions';
 import DialogTitle from '@material-ui/core/DialogTitle';
-import { Link } from 'react-router-dom';
+import { Link, useHistory } from 'react-router-dom';
 import axios from 'axios';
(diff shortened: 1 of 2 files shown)


client/src/components/More.jsx
```diff
@@ -105,6 +106,10 @@ const More = () => {
       });
   };
 
+  const navigateTo = href => {
+    history.push(href);
+  }
+
```
SomeMoosery, line 112: Any reasoning for switching to `useHistory`?

AlexRAV: Yeah. Because of we using Button component for these elements and for JSX react-router requires use Link component, as far as I know there are only two ways how to fix issue: restyle Link component to make it look like Button or just use Button component and on click programmaticaly redirect user on specified page. The second option is pretty natural and has no difference from using Link in terms of redirecting user on another page. It's just don't affects on styles unlike Link component.

SomeMoosery: Great - makes a lot of sense!

SomeMoosery:
Will just wait for tests to pass and then merge - looks good!

AdhamAH:
Neat! thanks!

AlexRAV:
Glad to help you)

PR #475, opened 2020-04-11
Rework redux blockstack

### LINKED ISSUE #457: Improve observation blockstack db interactions
## Summary

I think that there is an issue there around fetching observations as their number grows that might need addressing, and I've meant to bring it up for a long time.  

The way we do it currently we list all files, iterate over that list returning a list of all files whose name includes "observation/", and then fetch each of those files one by one.  

Similarly when we create a new observation we do this thing where we have to list and count all of the files in order to determine the name of the next file we create.  

I think that we could just have a single 'observation' file which contains a json string array of all the observation objects.  Fetching all the observations would just look like JSON.parse(usersession.getfile('observations')) and adding an observation would just mean adding the new observation to that array and userSession.putFile('observation', JSON.stringify(that new objservations array))

## Motivation

Turn these exponential operations into constant time. Really simplify all the blockstack db interactions.

fallon7284:

fixes #457

This PR touches a lot of files, changing the whole flow of how a survey form/observation is submitted to gaia db, how it is structured there, how it is fetched and persisted in redux.  It reduces complexity and improves code clarity.

We previously stored each model in a different file in gaia.  This made posting and fetching from db overly complex.

### Test plan (required)

### Final Checklist

- [ ] Have you squashed all commits, or if you will merge will you squash all commits?
- [ ] Did you use yarn, not npm?


acthelemann:
Is this a WIP? Or should I review it?

fallon7284:
@acthelemann no it's finished.

.env.development
```diff
@@ -0,0 +1,2 @@
+REACT_APP_QA_URL=http://ec2-54-92-173-146.compute-1.amazonaws.com:1260
+REACT_APP_MAPBOX_TOKEN =pk.eyJ1IjoiYXJpaHdhbmciLCJhIjoiY2s4NHRrdzd5MDExaTNtcGN4eGZnaGM3dCJ9.Vk4gts-RAG3GQxEIhXLg2g
```
SomeMoosery, line 2: should remove this `.env` file

client/src/components/App.jsx
```diff
@@ -18,11 +18,16 @@ import About from './About';
 import Disclaimer from './Disclaimer';
 import NotFoundPage from './NotFoundPage';
 import actions from '../redux/actions/actions';
+// import { fetchDemographicsComorbidities } from '../redux/actions/onboarding'
```
SomeMoosery, line 21: Remove this comment and others in file.

Also, why did you opt to remove this here?

client/src/redux/reducers/submit-survey.js
```diff
@@ -1,16 +1,16 @@
-export const SUBMIT_SURVEY = 'SUBMIT_SURVEY';
+// import { SUBMIT_SURVEY } from '../actions/submit-survey';
 
-const initialState = {
-  isSubmitted: null,
-};
+// const initialState = {
+//   isSubmitted: null,
+// };
```
SomeMoosery, line 5: remove comments in this file

client/src/tests/redux/reducers/observations.test.js
```diff
@@ -1,25 +1,28 @@
...
-    };
-    expect(reducer({}, numObs)).toEqual({ numObservations: 2 });
-  });
+  // it('should handle NUM_OBSERVATIONS', () => {
+  //   const numObs = {
+  //     type: NUM_OBSERVATIONS,
+  //     numObservations: 2,
+  //   };
+  //   expect(reducer({}, numObs)).toEqual({ numObservations: 2 });
+  // });
```
SomeMoosery, line 20: remove comments in this file

client/src/components/survey/SurveyPage3.jsx
```diff
@@ -104,8 +104,7 @@ const mapDispatchToProps = dispatch => {
   return {
     setSurveyPage3: survey => dispatch(actions.setSurveyPage3(survey)),
     toSurveyPage2: survey => dispatch(actions.toSurveyPage2(survey)),
-    submitSurvey: (fileNumber, observation, encryptOptions, userSession) =>
-      dispatch(actions.submitSurveyThunk(fileNumber, observation, encryptOptions, userSession)),
+    addObservation: (userSession, survey) => dispatch(actions.addObservation(userSession, survey)),
```
SomeMoosery, line 107: Unless I'm missing something, it looks like you changed the function that adds an observation from a thunk to a typical action - shouldn't we keep this as a thunk?

client/src/components/App.jsx
fallon7284: So I actually didn't remove this, I added it and then commented it -- my bad for leaving it there without addressing it, it was outside the scope of this issue but it was something I noticed in the course of working on this -- that we were adding this detail about the individual into gaia and then never using it, I wasn't sure if we also wanted that in redux so I wrote that thunk that we could add if we want that data available.

fallon7284: I can just remove it from this PR, I left the comment in more to remind myself.

client/src/components/survey/SurveyPage3.jsx
fallon7284: It's a thunk.

fallon7284:
So this is what the new observations.json looks like. the top half of this image is the console log of the response from gaia.  It's a json string array of the observations objects.  The bottom half of the image is the redux observation state after that json string is converted into a javascript array of objects.
Screen Shot 2020-04-12 at 8 36 48 AM

fallon7284:
And this is submitting a survey. You can see the two separate gaia uploads, one for observations.json and one for observation/0000004.json.  Below you can see that a forth observation has been added to the redux store.

Screen Shot 2020-04-12 at 8 43 34 AM

client/src/utils/blockstackHelpers.js
```diff
@@ -0,0 +1,18 @@
+const postObservation = async (userSession, observation, observationNumber) => {
+  const fileNumber = `${observationNumber}`.padStart(7, '0');
+  userSession.putFile(`observation/${fileNumber}.json`, JSON.stringify(observation));
+  const result = await userSession.putFile(`observations.json`, JSON.stringify(observation));
+  return result;
+};
```
acthelemann, line 6: This function is a bit unclear. I understand it, but I think it should cleared up.

- The second argument, `observation`, is really expecting an array of observations not just one
- The first `putFile` call should be putting an individual observation, not an array

Maybe separate putting an individual observation and the list of observations into different functions. Up to you.

client/src/redux/actions/observations.js
```diff
@@ -1,57 +1,33 @@
...
-    if (fileName.includes('observation/')) {
-      fileNames.push(fileName);
+import { getObservations, postObservation } from '../../utils/blockstackHelpers';
+import { clearSurvey } from './survey';
+
+export const ADD_OBSERVATION = 'ADD_OBSERVATION';
+export const FETCH_OBSERVATIONS = 'FETCH_OBSERVATIONS';
+
+export const addObservation = (userSession, observation) => async dispatch => {
+  const obs = await getObservations(userSession);
```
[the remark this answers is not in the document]
fallon7284: So I'm just going to quickly outline what goes on here just to be sure everyone is on the same page...

Because our observations have to be stored as strings in gaia, we need to get all the observations from there, parse them into a JS array, add another item to that array, and stringify them again.

I think that what you're saying is in line with that, and it is just that you're saying rather than check the db for the current list of observations, we should check redux because we should already have the observations which we fetch when we first initialize the application.  I did consider this and the reason why I thought it was worth the extra call to gaia was the consideration of like... what if, for some unexpected and unlikely reason, our initial gaia data fetch had failed and we don't have that list in redux?  In that case, if we just check redux, and the new observation, and post that list to redux, we will have overwritten the observations file in gaia.

So I thought this approach was safety, and that was worth the extra call to gaia.

Thoughts?

client/src/components/survey/SurveyPage3.jsx
SomeMoosery: Whoops, sorry about that

client/src/redux/actions/observations.js
```diff
@@ -1,57 +1,33 @@
...
-  });
-
-  dispatch(setNumObservations(numObservations));
-
-  // Convert the observations to JSON objects
-  (await Promise.all(getFilePromises)).forEach(observationString => {
-    const observation = JSON.parse(observationString);
-    observations.push(observation);
-  });
+  dispatch(clearSurvey());
```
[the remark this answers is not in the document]
fallon7284: _I feel the addObservation thunk should have nothing to do with a survey_

I'm not sure that I totally understand... an observation is the survey data.  The addObservation thunk is the submission of a survey.  So clear survey (not a function I wrote, I'm just using the actions that are already built to manage the survey ui state) would follow the submission of the survey.  

I can certainly see the argument for clarity sake of moving the clearSurvey outside the submit action and just call it directly after in the component.

client/src/redux/actions/observations.js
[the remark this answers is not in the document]
acthelemann: Makes sense. What do you think about checking Redux first, and if nothing is found, fetch from Gaia?

client/src/redux/actions/observations.js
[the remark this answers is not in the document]
acthelemann: Yeah, I guess it's just the naming that's throwing me off. I see that the only way to add an observation is to submit a survey, but when I think of a function `addObservation`, I wouldn't think it touches the survey, like with `clearSurvey`.

> I can certainly see the argument for clarity sake of moving the clearSurvey outside the submit action and just call it directly after in the component.

This is really what I'm getting at.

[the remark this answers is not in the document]
SomeMoosery: I agree with the above ^ good pointing out. If you're willing I think it would be worthwhile to move `clearSurvey` outside of this action, although we can make another issue for it as well.

client/src/utils/blockstackHelpers.js
fallon7284: _The second argument, observation, is really expecting an array of observations not just one_

No, the second argument is definitely only ever expecting one observation.  This function receives 1) The user session, 2) the form data from the survey, 3) the number of the next observation/00000x.json file, derived from the length of the array we fetched from observations.json

fallon7284: That's a good catch.  I wrote the whole observations.json and added the sort of backup post to create the old-style observations/00000x.json after.  You're right, I'm posting the whole array in both cases.  I will fix that.

client/src/redux/actions/observations.js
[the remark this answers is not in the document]
fallon7284: I'll move it outside the action, back to the component.

client/src/redux/actions/observations.js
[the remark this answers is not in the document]
fallon7284: I'm not sure that avoiding the one call to gaia seems with the added complexity?  But I'd be happy to hear others weigh in.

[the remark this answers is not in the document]
SomeMoosery: I think it would be fine for now the way it is, but as we hopefully grow it may be worth adding the complexity now to cut down on network requests where we can to keep things snappy

fallon7284:
so I've separated the gaia post actions into two different functions.  You can see what that for the observations.json we post the full json stringified array, and then when that is posted we post the single observation/000000x.json

Screen Shot 2020-04-12 at 1 01 37 PM
Screen Shot 2020-04-12 at 1 01 51 PM

client/src/redux/actions/observations.js
```diff
@@ -1,57 +1,31 @@
...
+  let obsArray;
+  let fileNumber = 1;
+  if (obs) {
+    const currentArray = JSON.parse(obs);
+    fileNumber = currentArray.length + 1;
+    obsArray = [...currentArray, observation];
+  } else {
+    obsArray = [observation];
+  }
+  postObservationsList(userSession, obsArray, fileNumber).then(didPost => {
```
acthelemann, line 17: Passing an extra argument here, but it works, no worries.

SomeMoosery:
Looks great - this is a really impactful longterm change that'll really help out going forward 🚀

PR #479, opened 2020-04-12
Refactors male & female gender buttons into dropdown list

### LINKED ISSUE #470: Update Gender Labels in Settings Screen
Currently, the only two options for genders are _Male_ and _Female_. We should either be adding a third button for _Non-Binary_ or, something that could be much more dynamic, turn it into a dropdown menu in case we add more in the future (e.g. genderqueer, gender-fluid, **trans male, **trans female). 

**The last two should only be added if we think that, for analytical purposes, knowing their sex-assigned at birth will be relevant. At the moment, it is not so we don't need to have that.

image

kevin-Price:
Refactors male & female gender buttons into dropdown list

Fixes mui components from overriding page css

fixes #470 

This fixes issue 470. Removes the Male and Female buttons and uses a muiSelect component instead. An extra gender "non-binary" has been added. 

The muiSelect component was using its own css instead of inheriting from the page (body) css. Therefore, I needed to override those css selectors in Layout.jsx. 

We want to be more gender inclusive, and provide an easy way to extend to more genders in the future. The muiSelect component gives us that ability.

### Test plan (required)

Before: 
Screen Shot 2020-04-12 at 2 02 03 PM
After: 
Screen Shot 2020-04-12 at 2 01 41 PM
Screen Shot 2020-04-12 at 2 03 20 PM

### Final Checklist

### DIFF
--- a/client/src/components/Layout.jsx
+++ b/client/src/components/Layout.jsx
@@ -13,7 +13,6 @@ const theme = createMuiTheme({
     MuiSelect: {
       select: {
         '&&': {
-          // overrides .MuiSelect-select.MuiSelect-select
           paddingRight: '0px',
         },
       },
--- a/client/src/components/Layout.jsx
+++ b/client/src/components/Layout.jsx
@@ -9,6 +9,21 @@ import NavBar from './NavBar';
 import Login from './Login';
 
 const theme = createMuiTheme({
+  overrides: {
+    MuiSelect: {
+      select: {
+        '&&': {
+          // overrides .MuiSelect-select.MuiSelect-select
+          paddingRight: '0px',
+        },
+      },
+    },
+    MuiInputBase: {
+      root: {
+        fontSize: 'inherit',
+      },
+    },
+  },
   palette: {
     primary: {
       light: '#c8d7fa',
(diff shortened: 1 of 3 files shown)


client/src/components/Layout.jsx
```diff
@@ -9,6 +9,21 @@ import NavBar from './NavBar';
 import Login from './Login';
 
 const theme = createMuiTheme({
+  overrides: {
+    MuiSelect: {
+      select: {
+        '&&': {
+          // overrides .MuiSelect-select.MuiSelect-select
```
SomeMoosery, line 16: remove comment

client/src/components/OnboardUser.jsx
```diff
@@ -145,40 +145,23 @@ const OnboardUser = props => {
...
+                  name="gender"
+                  value={formState.gender}
+                  onChange={handleChange}
+                  displayEmpty
+                  style={formState.gender !== '' ? { backgroundColor: '#82a4f8', outline: 'none', color: 'wheat' } : {}}
                 >
-                  Female
-                </button>
-              </ButtonGroup>
+                  <MenuItem value="" disabled>
```
SomeMoosery, line 157: Do you need to set `value` here at all?

kevin-Price: This does need to be here in order for the "click here" text to show.

client/src/components/Layout.jsx
kevin-Price: Ill make this change and resubmit!

PR #496, opened 2020-04-14
Feat/delete data

### LINKED ISSUE #476: Delete observation functionality
## Summary

We want to give users the ability to delete any observation they wish

**Acceptance Criteria** 

- [ ] Add ability to swipe right on a WeeklyTracker for a specific observation
- [ ] on right-swipe, implement a `Dialog` asking if the user is sure they want to delete, and then a redux thunk to delete that observation (using Gaia `deleteFile()`)
- [ ] update UI to reflect that deleted observation

## Motivation

Both to give users a sense of ownership over their data (since they do own it) and... for way down the line... to comply with GDPR guidelines that users should always be able to delete their data upon request.

None, this is necessary

JWaltuch:

fixes #476 

I added a button in the settings for a user to delete all data, which seems to make logical sense. A dialog will render asking if a user is sure they want to delete. If they click yes, it deletes the observation file, all individual observation files, and clears the detailData and observations in the redux store.

This should satisfy the legal requirements that users can delete their data at any time.

### Test plan (required)

gif

### Final Checklist

- [ x ] Have you bumped the version in `package.json`?
- [ x ] Have you added any new tests necessary?
- [ x ] Is your PR rebased off the most current master?
- [ x ] Have you squashed all commits, or if you will merge will you squash all commits?
- [ x ] Did you use yarn, not npm?
- [ x ] Did you use Material-UI wherever possible?
- [ x ] Did you format according to Prettier?
- [ x ] Did you run all of your most recent changes locally to make sure everything is working?


client/src/components/OnboardUser.jsx
```diff
@@ -96,6 +97,7 @@ const OnboardUser = props => {
   const { setDemographicsComorbiditiesThunk } = props;
   const { userSession } = useBlockstack();
   const [formState, setFormState] = useState(blankForm);
+  const [userWantsToDelete, setuserWantsToDelete] = useState(false);
```
SomeMoosery, line 100: This is nitpicky, only change if you feel the need, but maybe change the name to something a little less "action-y" - not sure else how to describe it. Something like `[deleteAllFiles, setDeleteAllFiles]`?

client/src/components/OnboardUser.jsx
```diff
@@ -340,6 +342,15 @@ const OnboardUser = props => {
...
           </Button>
         </div>
+        <Button
+          onClick={() => {
+            setuserWantsToDelete(true);
+          }}
+          className={classes.saveButton}
+        >
+          DELETE ALL OBSERVATION DATA
+        </Button>
```
SomeMoosery, line 352: Maybe we should make this button red to make sure users are hyper-aware of this?

SomeMoosery:
**For anyone reviewing** - take a look at the referenced ticket where we decided for now adding basic "delete all" functionality is a good first step leading up to Friday's beta launch, and from here we'll add functionality to delete specific files as how to best do that becomes a more involved conversation (modifying `observations.json`, dealing with currently basing adding new observations off of total observations, and if we delete one that will throw things off, etc)

client/src/utils/blockstackHelpers.js
```diff
@@ -8,6 +8,17 @@ export const postSingleObservation = async (userSession, observation, observatio
   userSession.putFile(`observation/${fileNumber}.json`, JSON.stringify(observation));
 };
 
+export const deleteAllObservations = async userSession => {
+  await userSession.deleteFile(`observations.json`);
+};
+
+export const deleteSingleObservations = async (userSession, numOfObservations) => {
+  for (let num = 1; num < numOfObservations; num += 1) {
```
acthelemann, line 16:
```suggestion
  for (let num = 1; num <= numOfObservations; num += 1) {
```

Make sure to delete the last observation.

client/src/redux/actions/observations.js
```diff
@@ -22,6 +29,18 @@ export const addObservation = (userSession, observation) => async dispatch => {
   });
 };
 
+export const deleteObservations = userSession => async dispatch => {
+  const obs = await getObservations(userSession);
+  if (obs) {
+    const obsArray = JSON.parse(obs);
+    const numOfObservations = obsArray.length - 1;
```
acthelemann, line 36:
```suggestion
    const numOfObservations = obsArray.length;
```

The length of `obsArray` is the number of observations you want to delete.

client/src/components/OnboardUser.jsx
acthelemann: I'll throw in another idea:

`[showDeletionDialog, setShowDeletionDialog]`

client/src/utils/blockstackHelpers.js
```diff
@@ -8,6 +8,17 @@ export const postSingleObservation = async (userSession, observation, observatio
   userSession.putFile(`observation/${fileNumber}.json`, JSON.stringify(observation));
 };
 
+export const deleteAllObservations = async userSession => {
+  await userSession.deleteFile(`observations.json`);
+};
+
+export const deleteSingleObservations = async (userSession, numOfObservations) => {
```
tylerd92, line 15: I would suggest renaming deleteSingleObservations to something like deleteObservation. Only to clearly demonstrate that the method only deletes one observation. When I first looked at the method I was not sure if it deleted only one observation.

client/src/redux/actions/observations.js
```diff
@@ -22,6 +29,18 @@ export const addObservation = (userSession, observation) => async dispatch => {
...
+    deleteAllObservations(userSession).then(() => {
+      deleteSingleObservations(userSession, numOfObservations);
```
Afrogrrrl, line 38: just a small detail here, make sure to put these kinds of statements into try/catch blocks. you can also do `.catch` at the end of all the  `.then` logic. And remember to put a status code to send back after each action (I've been using 200)

client/src/utils/blockstackHelpers.js
acthelemann: The method can delete more than one observation.

client/src/utils/blockstackHelpers.js
```diff
@@ -8,6 +8,17 @@ export const postSingleObservation = async (userSession, observation, observatio
   userSession.putFile(`observation/${fileNumber}.json`, JSON.stringify(observation));
 };
 
+export const deleteAllObservations = async userSession => {
+  await userSession.deleteFile(`observations.json`);
+};
```
Afrogrrrl, line 13: needs a try/catch and response code here too I think

client/src/redux/actions/observations.js
acthelemann: I'm personally confused about the status code stuff. Can you explain further or send me some documentation on this?

client/src/utils/blockstackHelpers.js
tylerd92: Ok I understand now that there is an observations file and there is a file for every single observation.

client/src/redux/actions/observations.js
Afrogrrrl: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/200 

Since we're sending a request to change data we want to make sure not only that the request is made but also that it is successful. If the request is successful it will continue on to the `.then` where we can signal with a status code (in this case 200 for OK). otherwise it will go into the `.catch` where we'll get a log of the error. I think most dBs will send these codes automatically so you don't have to hard code them in redux doesn't do it on it's own so I think it's good practice to put these things in, in the case that if one day this request fails it will be quicker to isolate the issue. If we're not getting a status of 200 we know the data wasn't changed in the way we requested.

fallon7284:
One question I would have is when we said "delete everything" did we mean delete all observations or delete all user data?  I was assuming it was the latter.  This leaves intact, for instance, the onboarding data.  Not sure what else is in gaia but at least that.

SomeMoosery:
I think for the scope of the story we should keep it to observations, but that's a good point, @fallon7284 

I can write out another issue for "Delete All User Data" - because that can still be very different and should have a different "are you sure?" type message. But, this does comply better with regulations (not in the US) giving individuals the right to be forgotten.

client/src/redux/actions/observations.js
Afrogrrrl: And thinking about this further, I'm curious if Gaia does have a response that it sends that we could just log here instead of hard coding it in?

acthelemann: I think the Blockstack API handles the HTTP-related stuff. We just focus on whether the `Promise` that the API call returns resolves or throws an error. I don't think we need to worry about handling status codes.

fallon7284:
Code looks pretty much on point except for the two bugs @acthelemann caught that need to be fixed to ensure that we actually delete all of the files -- 1) the number of files is the length of observations.json, not the length of observation.json - 1 and 2) the if you're looping from 1 to the number of observations, you're loop should run until <= numObservations.  Those are noted where needed and I'm going to make one more suggestion about naming functions on the relevant file, but otherwise looks good to me.

client/src/utils/blockstackHelpers.js
fallon7284: Yeah so one thought about naming, for the clarity of future contributors around what these functions actually do... delete all observations definitely doesn't delete all observations... delete all and delete single sound like sort of do different operations on the same list... like delete a whole list vs delete one item out of a list.  So I'd suggest something like deleteObservationsList and deleteObservationFile.  But I also think that naming is going to be inherently a little confusing here for anyone who doesn't realize in advance that we have this little bit of data redundancy going on.

client/src/redux/actions/observations.js
Afrogrrrl: cool, I'm still getting familiar with blockstack so I wasn't sure!

SomeMoosery:
Also, almost slipped my mind, let's add a couple of quick unit tests for the component and actions

client/src/redux/actions/observations.js
JWaltuch: I see the logic in the <= above, but since the array starts at index 0, don't I still want to subtract 1 here to get the correct number of observations?

acthelemann: The array does start at index 0, but the length of the array is the correct number of observations.

For example:

```js
const array = [ 1, 2, 3 ]; // This array starts at index 0,
console.log(array.length); // but its length is 3, which is how many numbers are in the array
```

Make sense?

client/src/utils/blockstackHelpers.js
JWaltuch: Hm... the method deletes many observations in a loop, so I worry deleteObservation also wouldn't be an accurate name?

client/src/redux/actions/observations.js
JWaltuch: Hehe yes thank you! Actually trying it out also showed me that you are correct.

JWaltuch:
Writing the tests will likely take me a good while, but so far I've got all the changes except the red button color. It's red but there's some property that makes it go transparent on hover that I'm trying to identify.

AdhamAH:
@JWaltuch Great work!
you can add to class `deleteButton` this:
```

    '&:hover': {
      boxShadow: '0px 1px 10px 0px #f64141',
      background: `#f64141`,
    },
```

client/src/redux/actions/observations.js
```diff
@@ -22,6 +29,23 @@ export const addObservation = (userSession, observation) => async dispatch => {
...
+    const obsArray = JSON.parse(obs);
+    const numOfObservations = obsArray.length;
+    deleteObservationsList(userSession)
+      .then(() => {
+        deleteObservationFiles(userSession, numOfObservations);
+        dispatch({ type: DELETE_OBSERVATIONS });
+        return 200;
+      })
+      .catch(err => {
+        throw new Error(err.message);
```
SomeMoosery, line 44: instead of throwing an error here, in other places we've opted to leave the `console.log` and override ESLint using `/* eslint-disable no-console */` at the top of the file

JWaltuch: Sure I can do that. And it depends how long it takes me to write the tests. I don't have much experience in that area and so far I have errors, but I intend to work on it today.

SomeMoosery: Sounds good! There's some similar examples already written out so definitely use those as reference. I would say, just in the essence of time as this is pretty important to include, if you can't figure out all the tests by tonight that we merge in this PR and write out another issue for testing.

JWaltuch: Works for me. And thanks! I'm working off the tests that exist so hopefully I can figure out the errors.

JWaltuch:
I think I got everything, so hopefully is ready to merge! :)

JWaltuch:
Haha I did bump it but I'm guessing when I merged in changes in bumped it back lol. Should be fixed!

JWaltuch:
Got it.

PR #609, opened 2020-04-21
Survey2 buttons

### LINKED ISSUE #586: Survey buttons de-select on mobile when scrolling
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:

### Description

When using CoronaTracker on an actual mobile phone (currently using iPhone X, both on Chrome and Safari) - filling out the survey page 2 and scrolling will cause buttons to become de-selected in the UI. They're still "selected" behind the scenes, but the UI shows them de-selected.

### Steps to Reproduce

1. Open the app on a mobile device
2. Take a survey

**Expected behavior:**

Buttons stay highlighted while scrolling

**Actual behavior:**

Buttons occasionally de-highlight while scrolling

**Reproduces how often:**

~80%

### Versions

1.0.0

### Additional Information

**This does not reproduce on the mobile view for Chrome dev tools, you need to test it on an actual mobile device**

PhurbaSherpa:

fixes #586 

Buttons would deselect in mobile browser. After inspecting it on mobile I found that the innerText obtained from clicking was different depending on where you clicked. If you log the e.target you'll see that clicking on the edge targets the button element but clicking in the middle targets the span element. So I inspected further and saw that the value for innerText were different, for the button it had a value of Minimal but with a new line vs the span where it has Minimal as we wanted. So when we do the ternary for determining a selected button it would be true if we clicked the middle but it would be false if we clicked the button. So instead I switched from using innerText to textContent so that either way the ternary would be true for clicking either. 

### Test plan (required)
1. Go on Mobile browser, and test it by purposely clicking the edge and watch it deselect after you click away. Pinch out to zoom in to accurately hit the outside.
2. Click the middle and you'll see it doesn't deselect. 

Clicking the outside gives you this:

button

Clicking Span gives you this:

span
 
Button textContent:

button2

Span textContent:

span2

### Final Checklist

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -1,6 +1,6 @@
 {
   "name": "corona-tracker",
-  "version": "1.0.9",
+  "version": "1.0.10",
   "homepage": ".",
   "private": true,
   "prettier": "@blockstack/prettier-config",
--- a/client/src/components/survey/SurveyPage2.jsx
+++ b/client/src/components/survey/SurveyPage2.jsx
@@ -271,7 +271,7 @@ const SurveyPage2 = props => {
               variant="contained"
               value="minimal"
               className={shortnessOfBreath === 'minimal' ? classes.selectedButton : classes.button}
-              onClick={e => handleshortnessOfBreath(e.target.innerText)}
+              onClick={e => handleshortnessOfBreath(e.target.textContent)}
             >
               Minimal
             </Button>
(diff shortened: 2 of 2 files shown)


SomeMoosery:
Nice catch! I'm just curious, which devices did you test on? Ideally, we want to be sure this will work on iOS as that seems to be second-most-common now

PhurbaSherpa:
> Nice catch! I'm just curious, which devices did you test on? Ideally, we want to be sure this will work on iOS as that seems to be second-most-common now

I used my iPhone 11 on verison 13.3.1

SomeMoosery:
Nice!

PR #612, opened 2020-04-21
Add CODEOWNERS

acthelemann:
Adds a `CODEOWNERS` file for the `/client/` directory.


acthelemann:
Just added the branch protection rules, so you'll see this PR needs two approving reviewers.

Admins can merge anyway, not sure if we want to not allow that.

SomeMoosery:
I would say let's keep admins as allowed to merge, just in case something accidentally breaks and we need to do a quick hotfix.

If it's not the case already I'd say we should just make you Adham and I admins for the time being (on mobile so hard to check myself)

acthelemann:
Yeah, I believe owners in the organization are admins by default. We are all owners.

https://github.com/orgs/COVID-19-electronic-health-system/people?query=role%3Aowner

acthelemann:
Any other people you think should be code owners? I'm not too familiar with other people

SomeMoosery:
I would also vote to add @pavel-ilin for this repo, thinking in terms of who contributes most often and for the longest amount of time.

We can expand the list as we continue 👍

AdhamAH:
I almost dropped my laptop!😅 I think I pressed by mistake the merge! I don't know what to say😂

acthelemann:
Haha should be fine

PR #613, opened 2020-04-21
Fix Login component - infinite spinning on close Blockstack modal

### LINKED ISSUE #611: Closing Get Started Dialog keeps you stuck at loading
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:

### Description

When initially coming to the app and clicking get started a popup shows up. If you close that pop up the app is stuck on the Loding component with no way of getting out except for refreshing.

loding

### Steps to Reproduce

1.  Start at Login page (sign out)
2.  Click Get Started
3.  Click the x on the top right

**Expected behavior:**
Probably should go back to the Login component

**Actual behavior:**

Is stuck on Loding component
**Reproduces how often:**
 100%

### Versions

1.0.10

### Additional Information

could remove the loding page since its only used there 
or
we could move setLoading(true) somewhere else instead of onclick of the 'Get Started' button since there is no way of setting loading to false after.

pavel-ilin:
fixes #611

Add event listener `onPointerOver` for parent `div` element in `Login` component. On event function `closeModal` invokes and reset value of the `loginLoading` variable. Which prevent infinite loading spin when we close Blockstack modal without login in.

### Final Checklist


acthelemann:
I checked out your branch and reproduced it:

spinz

pavel-ilin:
You are right about mobile version. Can't think about anything at the moment. It's tricky because we not control the Blockstack modal.

Will do more research.

acthelemann:
Hmm I'd hope there's something in the Blockstack API to tell when this happens.

If not, I'm not so sure that this loading screen is even necessary, but I'd like some others' thoughts before removing it. @AdhamAH @SomeMoosery

SomeMoosery:
I've been thinking about this, and was thinking maybe we only display it on desktop (or we could remove it entirely, frankly).

On mobile, it adds nothing, and on desktop the loading icon is hidden behind the Blockstack modal which serves to just hide the "Get Started" screen.

I'd honestly be fine removing it altogether, it seems to be more headache than it's worth

pavel-ilin:
I think we can remove it. It's not add aesthetic or functional value. If everyone's agree, I can do it.

SomeMoosery:
Yeah, let's send it 🏂 we can just keep removing it in this PR, and it'll close out the issue 👍

client/src/components/Login.jsx
```diff
@@ -55,9 +55,12 @@ const Login = props => {
     setLoading(true);
     doOpenAuth();
   };
+  const closeModal = () => {
+    setLoading(false);
+  };
```
SomeMoosery, line 60: I think `setLoading` only deals with the loading component itself, right? We could probably just remove this functionality entirely (from Redux as well) if we're going to remove the loading screen.

pavel-ilin: All references `setLoading`  removed from the app.

pavel-ilin:
> looks good but you need to update the package.json, sorry I just merged a PR

Version updated.

Issue #665, opened 2020-04-29
[FEAT] Python scripting: format GSheet csv to json via Google API data pull


ngiangre:
# ⚠️ IMPORTANT: Please fill out this template to give us as much information as possible to consider/implement the feature.

### Prerequisites

- [X] check this box if you have completed the following:
    - Reviewed the [contributing guidelines](https://github.com/COVID-19-electronic-health-system/.github/blob/master/CONTRIBUTING.md) and [support](https://github.com/COVID-19-electronic-health-system/.github/blob/master/SUPPORT.md) files
    - Reviewed the README file for the repository you are working in
    - Searched for relevant instructions [on our Discord server](https://discord.gg/pPERUuv)
    - Searched the issues of the repository you are working in to make sure one was not already filed

### Summary

This is a modular issue sprouting from #643 

All the translations are in an accurate format and now we need to pull the sheets (English, Spanish, French, Italian, Dutch (Netherlands), Russian) via the google API into the respository for front end developers to reference by the json keys.

There is a link in the python script `src/python/pull_gsheet_data.py` for creating your own google api key. This python script is just a start and needs more development. 

each translation sheet has parentKey, childKey, fieldKey, value, translatedValue, and then other columns. We need a structure of { 'parentKey' : { 'childKey' : { 'fieldKey' : { 'value' : '', 'translatedValue' : '', ... } } } }. In the case of education, { 'parentKey' : { 'childKey' : [ {'value': '', ... }, {'value' : '', ... }, ... ] } } 

The Date attributes are not needed - filter those out. 

The resulting json files should go into `public/locales/` though I put them in `docs/content` to not mess things up. 

Here's some extra code I already started using that might be helpful:
`
# Init sheet names and output dir #

data_model_sheet_name = "Data Model"
education_sheet_name="Education"
health_sheet_name = "Health"
translation_sheets_regex = " - Master Sheet"
translation_sheets_not_regex = "OLD"
languages_to_pull = ['English','Dutch (Netherlands)','Spanish','Italian','French','Russian']
out_dir = "../../docs/content/"

# Set dictionary to connect languages to two-letter abbreviations #
#https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes
language_letters_dict= { 'English' : 'en', 'Dutch' : 'nl', 'Spanish' : 'es',
                         'Italian' : 'it', 'French' : 'fr', 'Russian' : 'ru'}
`

### Motivation

[...]

PR #691, opened 2020-05-06
added nlp tokenizer to python gsheet pull script

ngiangre:
# ⚠️ __IMPORTANT: Please do not create a Pull Request without creating an Issue first.__

All changes need to be discussed before proceeding. Failure to do so may result in the pull request being rejected.

Before submitting a pull request, please be sure to review:
- Our Code of Conduct: https://github.com/COVID-19-electronic-health-system/.github/blob/master/CODE_OF_CONDUCT.md
- Our Contributing Guidelines: https://github.com/COVID-19-electronic-health-system/.github/blob/master/CONTRIBUTING.md
- Our Support documentation: https://github.com/COVID-19-electronic-health-system/.github/blob/master/SUPPORT.md

-----

## Please include the issue number the pull request fixes by replacing YOUR-ISSUE-HERE in the text below.

Fixes #YOUR-ISSUE-HERE

## Summary

This is in reference to #665 

This PR submits a python script for the mvp which pulls the gsheet content for translations and formats them accurately.

The translation.json files are added to the client/[language]/ folders.

## Final Checklist

- [ ] For CoronaTracker, did you bump the version in `package.json`?
    - Update the second decimal for a major change
    - Update the third decimal for a minor change
    - Numbers can go past 9, e.g. `1.0.9` => `1.0.10`
    - For more info, read about [Semantic Versioning](https://semver.org/)
- [ ] Did you add any new tests as necessary?
- [ ] Is your PR rebased off the most current master?
- [ ] Have you squashed all commits? _(can be done at merge)_
- [ ] Did you use `yarn` not `npm`? _(important!)_
- [ ] Did you use Material-UI wherever possible?
- [ ] Did you run `yarn lint` on the code?
- [ ] Did you run all of your most recent changes locally to make sure everything is working?


src/python/pull_gsheets_translations_mvp.py
```diff
@@ -0,0 +1,136 @@
...
+"""
+
+def convert_to_camelCase(value):
+	#converts value to camelCase
+	camelCase = value.split()[0].lower() + " ".join(value.split()[1:]).title().replace(" ","")
+	text_tokens = word_tokenize(text)
+	#removes punctuation
+	return camelCase.translate(str.maketrans('', '', string.punctuation))
+
+import nltk
```
SomeMoosery, line 72: This is just stylistic and I'm not a Python expert so correct me if I'm wrong, but I feel it's best practice to keep all imports at the top of a file

src/python/pull_gsheets_translations_mvp.py
```diff
@@ -0,0 +1,136 @@
...
+nltk.download('punkt')
+nltk.download('stopwords')
+from nltk.tokenize import word_tokenize
```
SomeMoosery, line 75: Same thing here

src/python/pull_gsheets_translations_mvp.py
```diff
@@ -0,0 +1,136 @@
...
+				for fieldKey,fgrp in cgrp.groupby(['fieldKey']):
+					fgrp_sub = fgrp.filter(regex='[Vv]alue')
+					valueDict = {"array":[]}
+					for value, translatedValue in fgrp_sub.values:
+						valueDict[value]=translatedValue
+						valueDict["array"].append(translatedValue)
+					fieldKeyDict[fieldKey] = valueDict
+				childKeyDict[childKey] = fieldKeyDict
+			parentKeyDict[parentKey] =  childKeyDict
+		wkDict.update(parentKeyDict)
```
SomeMoosery, line 132: Is there no python/pandas/numpy function that would allow us to avoid this structure? If not, it's probably fine for the time being but this many nested for loops can lead to nasty worst case runtime complexity so it just stood out to me

SomeMoosery: [this may be a good reference](https://stackoverflow.com/questions/42564432/optimizing-a-quadruple-nested-for-loop)

ngiangre: Thanks for the reference! 

I’m going to look into the style for imports in a script and then do another commit. 

Right now the nested loops are needed. I was using pandas ‘to_json’ but it isn’t specific enough to produce both the array and value keys. I think we could do more effective list comprehensions but right now there shouldn’t be a concern for complexity since there’s so few rows in the gsheets. But definitely in the future we need better checks for a potential runtime error. 

Also is there a requirements.txt or something to require the python library dependencies?

SomeMoosery: Fair enough - I figured this may be unavoidable for now, and even so it won't make much of a difference since `O(n^2)` vs `O(n^4)` is likely a matter of milliseconds when it's not thousands or hundreds of thousands of rows lol

And if you're asking if we can put in a `requirements.txt` I think we can just put it right in the `src/python` directory right? I honestly have very limited experience with python and have never used it in a professional setting. Feel free to add a `requirements.txt` to this PR, or if we want to do that in a separate PR at least add the required libs to the `README` for contributors to get up and running - good catch

ngiangre: No great point - for scalability we need to reduce the complexity. But I’m not the best at doing so I’ll let a software engineer do that :)

I’m used to ‘pip install requirements.txt’ for getting libraries installed locally from modules or repos so I’ll include a requirements.txt in the PR soon. But again I do not know the most efficient way to mix js and python repos so in the future this should be amenable. Maybe we can have bc a feature enhancement issue?

SomeMoosery: Yeah I'd say for now just adding the `requirements.txt` in `src/python` should be fine with no overlap 👍

ngiangre: added it to the top of the dorectory - I don't ever see it in src/ folders

src/python/pull_gsheets_translations_mvp.py
ngiangre: changed to top of script

src/python/pull_gsheets_translations_mvp.py
ngiangre: done

src/python/pull_gsheets_translations_mvp.py
acthelemann: I think it's better in `src/python` for organization

I think [pipenv](https://pipenv.pypa.io/en/latest/) has replaced `requirements.txt`, but we can handle that in a future issue/PR

acthelemann:
A basic readme on a general overview of what this does and how to run it in `src/python` would be helpful

src/python/pull_gsheets_translations_mvp.py
```diff
@@ -0,0 +1,299 @@
...
+
+	Returns:
+		[string] -- [returns a string with no punctuation, in camelCase format.]
+	"""
+	# if value is a string with more than one word
+	if len(value.split()) > 1:
+		#converts value to camelCase
+		camelCase = value.split()[0].lower() + " ".join(value.split()[1:]).title().replace(" ","")
+		#removes punctuation
+		return camelCase.translate(str.maketrans('', '', punctuation))
```
acthelemann, line 65: For consistency, let's pick one depunctuation method, `translate(str.maketrans('', '', punctuation))` or `depunctuate`, and use it everywhere depunctuation is done

src/python/pull_gsheets_translations_mvp.py
```diff
@@ -0,0 +1,299 @@
...
+		nouns = [depunctuate(x[0]) for x in pt if "NN" in x[1]]
+		#choose verbs in setence
+		verbs = [depunctuate(x[0]) for x in pt if 'VB' in x[1]]
+		#make fieldKey and add to new list of fieldKeys
+		if len(verbs)==0 and len(nouns)==2:
+			lst_new.append(nouns[0].lower() + nouns[1].title())
+		elif len(verbs)==1 and len(nouns)==1:
+			lst_new.append(nouns[0].lower() + verbs[0].title())
+		else:
+			lst_new.append(nouns[0].lower() + nouns[1].title() + verbs[0].title())
```
acthelemann, line 107: This is a bit fragile, but if it works with the current sheets it should be fine. We can fix it later

src/python/pull_gsheets_translations_mvp.py
```diff
@@ -0,0 +1,299 @@
...
+		orig_row = row.copy()
+		orig_row[v_col] = convert_to_camelCase(orig_row[v_col])
+		rows_added.append(orig_row)
+		arr = row[vPA_col].split(';')
+		tarr = row[tVPA_col].split(';')
+		if len(arr)==len(tarr) and len(arr)>3:
+			for j,_ in enumerate(arr):
+				row_new = row.copy()
+				row_new[v_col] = arr[j].replace('\n','')
+				row_new[v_col] = convert_to_camelCase(row_new[v_col]).replace(" ","")
```
acthelemann, line 135:
```suggestion
				row_new[v_col] = convert_to_camelCase(row_new[v_col])
```

`convert_to_camelCase` does the `replace` I believe

src/python/pull_gsheets_translations_mvp.py
```diff
@@ -0,0 +1,299 @@
...
+		df {[DataFrame]} -- [The current language dataframe.]
+
+	Returns:
+		[DataFrame] -- [A cleaned dataframe with the necessary columns for the JSON key:value pairs.]
+	"""
+	if 'Education' in wk.title:
+		df.value = education_value_cleaner(df)
+	elif 'Survey' in wk.title:
+		osLanguage_df = df
+		df = survey_value_cleaner(df)
```
acthelemann, line 170: It would be more consistent if `education_value_cleaner` and `survey_value_cleaner` returned the same type

src/python/pull_gsheets_translations_mvp.py
```diff
@@ -0,0 +1,299 @@
...
+						valueDict = {"array":[]}
+						for value, translatedValue in fgrp_sub.values:
+							valueDict[value]=translatedValue
+							valueDict["array"].append(translatedValue.replace(' ','',1))
+						fieldKeyDict[fieldKey] = valueDict
+					childKeyDict[childKey] = fieldKeyDict
+				parentKeyDict[parentKey] =  childKeyDict
+			wkDict.update(parentKeyDict)
+
+		save_to_JSON(OUT_DIR,locale,wkDict)
```
acthelemann, line 298: Can we call `save_to_JSON` once per `language` instead of once per `wk`?

ngiangre:
you're totally right @acthelemann these are great suggestions and should be made. I can make another issue for this - as a feature enhancement. I think we need to move to the next step of working with these translations so the front end devs can try to use it. Getting feedback should be prioritized if since we'll most likely need to change functions/methods from that as well.

